PR #3, opened 2018-04-18
Add Patternfly-React

aljesusg:
Add patternfly-react to the project

What do you think about this configuration @dmiller9911 ? I added and alert success to the Home

a


webpack.config.js
```diff
@@ -39,6 +45,24 @@ module.exports = env => {
...
+          test: /\.(css|scss|sass)$/i,
+          use: [
+            {
+              loader: 'style-loader'
+            },
+            {
+              loader: 'css-loader'
+            },
+            {
+              loader: 'sass-loader'
```
dmiller9911: I am thinking we will not need this.  This project will be using glamorous for styles that do not come from Patternfly:  https://glamorous.rocks/

webpack.config.js
```diff
@@ -39,6 +45,24 @@ module.exports = env => {
         {
           test: /\.html?$/,
           use: [{ loader: 'html-loader' }]
+        },
+        {
+          test: /\.(css|scss|sass)$/i,
+          use: [
+            {
+              loader: 'style-loader'
```
dmiller9911: Let's use this: https://github.com/webpack-contrib/mini-css-extract-plugin#configuration. It will allow us to code split css later. We will also want to look at minimizing for production too: https://github.com/webpack-contrib/mini-css-extract-plugin#minimizing-for-production, but that can wait for later as well since I believe Patternfly CSS is already minified.

package.json
```diff
@@ -58,18 +60,24 @@
...
     "jest": "^22.4.3",
     "jest-glamor-react": "^4.2.1",
+    "node-sass": "^4.8.3",
+    "patternfly-react": "^1.19.1",
     "react": "^16.3.1",
     "react-dom": "^16.3.1",
     "react-hot-loader": "^4.0.1",
     "react-i18next": "^7.5.1",
     "react-router-dom": "^4.2.2",
+    "react-svg-loader": "^2.1.0",
```
dmiller9911: This does not look to be used yet. Let's remove it for now, and we can add it with its use.

webpack.config.js
aljesusg: This comment go down ^^. I think that there is a min version for patternfly.

webpack.config.js
aljesusg: I need to check it again because yarn start throw errors about to add a loader to handle this type of files, I think that patternfly-react css import them

dmiller9911: css shouldn't need sass-loader.  We do need style-loader if we do not use the css plugin.  An example of the config for the css plugin would look like this:
```javascript
// webpack.config.js
module.exports = () => {
  return {
    ...rest of config,
    module: {
       rules: [
         ...rest of rules,
         {
            test: /\.css$/,
            use: [MiniCssExtractPlugin.loader, 'css-loader']
         },
       ]
    },
    plugins: [
        ...rest of plugins,
        new MiniCssExtractPlugin({
           filename: '[name].css',
           chunkFilename: '[id].css'
        })
    ],
  }
}

```

aljesusg:
@dmiller9911 Ok all is running with changes, about the RCUE I don't know if you wanna depend of RCUE to upgrade patternfly-react.

PR #53, opened 2018-07-25
Deploy on openshift

chambridge:
Updates to enable deployment onto Openshift with nginx.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -10,9 +10,7 @@ import { configureStore } from './store';
 initApi({
   host:
     process.env.NODE_ENV === 'production'
-      ? 'http://koku-' +
-        process.env.APP_NAMESPACE +
-        '.1b13.insights.openshiftapps.com/'
+      ? 'http://koku-koku-dev.1b13.insights.openshiftapps.com'
       : 'http://localhost:8000',
   version: 'v1',
 });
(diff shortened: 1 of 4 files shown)


dmiller9911:

Built with commit 811e589aace05d1023ec8a6327a144da50aa9d6c

https://deploy-preview-53--koku.netlify.com

src/index.tsx
```diff
@@ -10,7 +10,7 @@ import { configureStore } from './store';
 initApi({
   host:
     process.env.NODE_ENV === 'production'
-      ? 'http://koku-koku.1b13.insights.openshiftapps.com/'
+      ? 'http://koku-koku-dev.1b13.insights.openshiftapps.com'
```
chambridge: Tried utilizing `process.env.APP_NAMESPACE` instead of hard-coding `koku-dev` here but it didn't seem to pick up the environment variable from from the ConfigMap.

image


image

I figure we could follow up with a secondary PR to get this so it will work in staging too.

dmiller9911: We have to explicitly add environment variables to the client.  We can update that in another PR.

chambridge:
@dmiller9911 like the following:

```
if not DEBUG:
    CORS_ORIGIN_REGEX_WHITELIST = (r'^(http(s)?}?://)?koku-ui-([a-zA-Z0-9_-]*\.)?1b13\.insights\.openshiftapps\.com$', )
```

PR #54, opened 2018-07-26
Use APP_NAMESPACE if it is available

dmiller9911:


dmiller9911:

Built with commit f1de640c7e79330d203f90d4bf97e7edacb947ed

https://deploy-preview-54--koku.netlify.com

src/index.tsx
```diff
@@ -8,10 +8,7 @@ import App from './app';
 import { configureStore } from './store';
 
 initApi({
-  host:
-    process.env.NODE_ENV === 'production'
-      ? 'http://koku-koku-dev.1b13.insights.openshiftapps.com'
-      : 'http://localhost:8000',
+  host: process.env.APP_NAMESPACE || 'http://localhost:8000',
```
chambridge: So `APP_NAMESPACE` is actually related to the Openshift project `koku-dev`, `koku-staging`, etc. So its a component of the url:
`'http://koku-' + process.env.APP_NAMESPACE + '.1b13.insights.openshiftapps.com'`

Let me know if that doesn't make sense.

dmiller9911: ah makes sense.  Updating. Will `https` vs `http` be an env setting in the near future?

PR #66, opened 2018-08-05
Handle null units from the api

dmiller9911:
The api can at times return null for the units in total for reports. Add a null check and return value if that is the case.


dmiller9911:

Built with commit f82f24265480ce5826b19bff088282be63095724

https://deploy-preview-66--koku.netlify.com

Issue #55, opened 2018-07-30
Login does not report errors to the user


chargio:
Describe the bug
Login does not report errors to the user.
Tried using a Mac on Firefox, Safari and Chrome.

To Reproduce

    go to login page
    Introduce login
    introduce password (bad password) 
    press the log in button / tab and press enter

Or if API is down time out should cause and error

Expected behavior
The user should get an error.

Context (please complete the following information):

    OS: MacOS X High Sierra
    Browser: chrome, safari, firefox
    Using staging app

chargio:
Additional comments on:

https://github.com/project-koku/koku/issues/283

dmiller9911:
@dlabrecq This would be a good one for you to pick up.  The error and loading status for the login page are in the redux store, but nothing is using them. I did not spend the time on this before since the insights login will take over in the future, but it looks like we will need a more robust login page. 

A Couple things we need to address:

1. Display an error for failed attempts.
1. Disable the submit button while the login is loading.  This will give some feedback to the user when they hit login.

PR #68, opened 2018-08-09
fix(login): add login page error handling

dlabrecq:
Wanted to add this PR now so @ripcurld can review.

Note this includes a temporary workaround for a type error with the pf4 alert.

Fixes https://github.com/project-koku/koku-ui/issues/55

screen shot 2018-08-09 at 10 09 40 am

### DIFF
--- a/src/pages/login/login.test.tsx
+++ b/src/pages/login/login.test.tsx
@@ -1,3 +1,4 @@
+import { AxiosError, AxiosResponse } from 'axios';
 import { TextInput } from 'components/textInput';
 import { shallow, ShallowWrapper } from 'enzyme';
 import React from 'react';
@@ -5,8 +6,8 @@ import { FetchStatus } from 'store/common';
 import { Login, Props } from './login';
 
 const props: Props = {
-  login: jest.fn(),
   fetchStatus: FetchStatus.none,
+  login: jest.fn(),
   t: jest.fn(v => v),
 };
 
@@ -37,6 +38,29 @@ test('triggers login on form submit', () => {
     username,
     password,
   });
+});
+
+test('triggers login error on form submit', () => {
+  const mockAxiosResponse: AxiosResponse = {
+    data: {
+      non_field_errors: 'Unable to log in with provided credentials.',
+    },
+    status: 1,
+    statusText: '',
+    headers: {},
+    config: {},
+  };
+  const mockAxiosError: AxiosError = {
+    config: {},
+    message: '',
+    name: '',
+    response: mockAxiosResponse,
+  };
+  const view = shallow(<Login {...props} error={mockAxiosError} />);
+  const preventDefault = jest.fn();
+  view.find('form').simulate('submit', { preventDefault });
+  expect(props.login).toBeCalled();
+  expect(view.find('.pf-c-alert')).not.toBeNull();
 });
 
 function getUsernameInput(view: ShallowWrapper<any, any>) {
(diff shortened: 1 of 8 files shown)


src/pages/login/login.test.tsx
```diff
@@ -39,6 +40,29 @@ test('triggers login on form submit', () => {
...
+    config: {},
+    message: '',
+    name: '',
+    response: mockAxiosResponse,
+  };
+  const view = shallow(<Login {...props} error={mockAxiosError} />);
+  const preventDefault = jest.fn();
+  view.find('form').simulate('submit', { preventDefault });
+  expect(props.login).toBeCalled();
+  expect(view.find('.pf-c-alert')).not.toBeNull();
```
dmiller9911: You can actually do this: 
```
import {  Alert } from '@patternfly/react-core';
...
expect(view.find(Alert).exists()).toBe(true);
```
While I am sure it is working currently the`.not.toBeNull()` is actually not a valid check here.  enzyme returns a collection no matter the result of finding something.  If it isn't there was nothing found it's length is just 0.

src/pages/login/login.test.tsx
```diff
@@ -39,6 +40,29 @@ test('triggers login on form submit', () => {
...
+  };
+  const mockAxiosError: AxiosError = {
+    config: {},
+    message: '',
+    name: '',
+    response: mockAxiosResponse,
+  };
+  const view = shallow(<Login {...props} error={mockAxiosError} />);
+  const preventDefault = jest.fn();
+  view.find('form').simulate('submit', { preventDefault });
```
dmiller9911: This is actually not needed here. The error will be there regardless since you passed it in as props. The login call is covered in another test.

src/pages/login/login.tsx
```diff
@@ -99,6 +122,9 @@ export class Login extends React.Component<Props, State> {
                   <div>
                     <Button
                       {...getTestProps(testIds.login.submit)}
+                      isDisabled={
```
dmiller9911: Would be good to add a test for this.
```jsx
test('Button is disabled if fetch status if in progress', () => {
  const view = shallow(<Login {...props} fetchStatus={FetchStatus.inProgress} />);
  expect(view.find(Button).props().isDisabled).toBe(true)
})
```

src/pages/login/login.test.tsx
dmiller9911: Additionally it might be nice to assert the text is correct. A shortcut for this would be
```
expect(view.find(Alert)).toMatchSnapshot();
```
That will verify the correct alert type, the correct testid and the text.

dlabrecq:
Added a red border color to the input field, similar to the PatternFly pattern. However, PF4 doesn't appear to have styles for that yet, so I improvised. I'll need Doug's input here regarding PF4 styling.

screen shot 2018-08-10 at 5 20 44 pm

src/pages/login/login.test.tsx
dlabrecq: Thanks, I actually had something similar to that, but didn't work properly because I wasn't using `.exists`

src/pages/login/login.tsx
```diff
@@ -90,6 +118,11 @@ export class Login extends React.Component<Props, State> {
                   <FormGroup label={t('login.password_label')}>
                     <TextInput
                       {...getTestProps(testIds.login.password_input)}
+                      {...(error &&
```
dmiller9911: This can be written as `isError={error && error.response && error.response.data.password}` if the typings ask for a boolean you can strictly cast it with `Boolean(error && error.response && error.response.data.password)` Same for the error above. It might be nice to move this up and assign them variable names something like
```jsx
const passwordError = error && error.response && error.response.data.password;
const usernameError = error && error.response && error.response.data.username;

<Alert>{usernameError || passwordError}</Alert>
<TextInput  isError={Boolean(usernameError)} />
<TextInput  isError={Boolean(passwordError)} />
```

That way you only need to dive into is once for each error.

PR #97, opened 2018-09-02
Added providers modal

dlabrecq:
Added modal to add providers. Although we were not able to use the PF3 wizard from frontigrade, we used the PF4 modal. This will be replaced when the PF4 wizard is available.

The modal is designed in a way to be opened from any page. We're currently only opening this from the dashboard, but believe we may launch from the cost details as well.

When the user logs in, the app fetches a list of providers which is saved via the Redux store. If there are no providers, the dashboard is replaced with a simple empty state asking the user to add an account.

When the user launches the providers modal, they are presented with 3 input fields; provider name, a bucket name, and the provider resource name. 

I've added some basic input validation; for example, to check if the bucket name has special characters, spaces, and is less than 255 chars. I also check the provider resource name to ensure  it begins with `arn:aws:`.

Empty state:

Providers Modal:

Modal with error checking:


dmiller9911:

Built with commit 4210786bb0acab751ec609c6c197a1ddcf3686c5

https://deploy-preview-97--koku.netlify.com

dlabrecq:

Built with commit 4210786bb0acab751ec609c6c197a1ddcf3686c5

https://deploy-preview-97--project-koku.netlify.com

PR #105, opened 2018-09-06
Add unit tests to tabs

boaz1337:


dlabrecq:

Built with commit 97979bdf8867a9f5b3e67c9051e7434017e34dd5

https://deploy-preview-105--project-koku.netlify.com

boaz1337:
Yeah, I thought about it too. 
Let's speak about it offline.

boaz1337:
@dlabrecq - I hope I didn't get too far but I think I found a better way to test without snapshot. WDYT?

dlabrecq:
I'm all for a better way. My thought was we should avoid relying on snapshots and use more traditional unit testing.

boaz1337:
great then you probably will love this change (no snapshot, only verifying through `props` and `text`). would you mind reviewing it?

PR #108, opened 2018-09-06
fix(Login): Alert title is not string

boaz1337:
Apparently the API returns the error data as an array.
For example, if `username` is missing the following error is returned:

```js
{
 ...
 "data": {
  ...
  "username": ["This field may not be blank"]
 }
}
```

This array is passed to the `title` props in `Alert` but it is supposed to be a string which ends up with printing a warning message in the console.

This commit fixes this by converting the array to string by using the `String` method.


dlabrecq:

Built with commit ccb05ca02440447cbdf1261d2989542b091ee5be

https://deploy-preview-108--project-koku.netlify.com

PR #125, opened 2018-09-17
fix(dashboardWidget): display daily data in tabs

### LINKED ISSUE #114: Charts are not using correct data
Charts should use the API for monthly data, while "top services", "top accounts", and "top regions" should use an API for daily data. Instead of seeing several (9+) services, the daily data should be limited to the top 5 most used. 

Some investigation is needed, but it's suspected that "top services" may be using the same API used to create the monthly chart?

screen shot 2018-09-12 at 2 21 05 pm

boaz1337:
The data that is rendered in tabs is taken from the current computed results which its resolution is set to 'daily' while it should be 'monthly'.

In this commit, another API call is done to fetch this data with the right parameters and render tabs according to it.

closes #114

### DIFF
--- a/src/api/query.ts
+++ b/src/api/query.ts
@@ -1,6 +1,6 @@
 import { parse, stringify } from 'qs';
 
-interface Filters {
+export interface Filters {
   time_scope_value?: number;
   time_scope_units?: 'month' | 'day';
   resolution?: 'daily' | 'monthly';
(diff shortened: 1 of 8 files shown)


dlabrecq:

Built with commit 6b07dceae857638a325d2e6bde42a3aab682177f

https://deploy-preview-125--project-koku.netlify.com

boaz1337:
Yep, no problem.

boaz1337:
I added more tests to `getComputedReportItems` to get 100% coverage.

chambridge:
Yes, the problem I mentioned to @dlabrecq was that the Top Services/Accounts/Regions info are using reports with the wrong query parameters. They are using the same report as the trendlines which pull back resolution of `daily` but these Top 5 items should be `monthly` resolution reports.

boaz1337:
:sunglasses: thanks a lot @chambridge I know how to fix this.

boaz1337:
@chambridge can you review this

PR #126, opened 2018-09-17
Added filter results to cost details toolbar

dlabrecq:
Added filter results to cost details toolbar

- This handles the case where multiple filters are applied.
- This handles the case where the page is refreshed or navigated away form.
- This handles the case where "Cost By" selection menu changes the page contents.
- This introduces patternfly-additions.css, which requires modifications to the existing styles created for the PF3 ListView.


dlabrecq:

Built with commit 5699e6bffdb8451cd3f431f2ceedb837d07a840a

https://deploy-preview-126--project-koku.netlify.com

andresgalante:
Hey @dlabrecq you might want to check if the new styles here #129 are ok with this PR

dlabrecq:
@andresgalante This PR introduces patternfly-additions.css in order to display the filter results in the toolbar. Many of the style overrides for ListView have been removed.

dlabrecq:
The previous filter field had `account`, `region`, & `service` as options. However, there is already a `Cost By` (group_by) selection which changes the data shown in the list. Applying a filter to `service` doesn't make sense when the `Cost By` menu is showing data for `accounts`. 

In order for the previous filter to work properly, `Cost By` must be the same choice as the filter field. When you first enter the Cost Details page, they are both set on `account` by default. Change the filter field so it does not match the `Cost By` (group_by) selection and the filter does not match.

The filter field should really have options such as `account`, `cost`, & possibly `cost delta` -- similar to what the sort is applies to. Note that when there is only one filter field option, the PF3 filter only shows an input and omits the drop down.

dlabrecq:
I'll demo the filter functionality for @nlcwong and make changes in another PR, if necessary.

dlabrecq:
I demoed the filter functionality for @nlcwong and she agreed this is what was intended for the POC.

PR #135, opened 2018-09-21
fix(packages): install necessary packages in production

### LINKED ISSUE #131: RFE: Move packages used for dev testing from required dependencies to devDependencies
## User Story
There are packages that are used for testing (jest-emotion, jest, enzyme,...) but marked as required for a production build of koku-ui.  If these packages are used exclusively in development then they should be marked as devDependencies and not installed by yarn during a production build (yarn --production)

See bug #127 

## Assumptions
- yarn is used as the package manager during the production build process

## Acceptance Criteria
No test packages are included in a production build of the UI

boaz1337:
fixes #131 

`enzyme`, `jest` and `webpack-dev-server` are needed only in development and thus should not be installed when running in production.
Currently, these packages are installed when building koku UI.
To fix this, these packages were moved from `dependencies` to `devDependencies` at `package.json`.
In addition, setting `env` to production instead of development in the `build:client` script.

I tested this by running yarn container:test and verifying the container didn't have errors and was running.
Also made sure the dev environment is still working and nothing is broken.

// ping @LandoCalrizzian


package.json
```diff
@@ -9,7 +9,7 @@
   "private": true,
   "scripts": {
     "build": "yarn build:client",
-    "build:client": "yarn clean && webpack --env development",
+    "build:client": "yarn clean && webpack --env production",
```
chambridge: I had to set this to `development` and make alterations to the Dockerfile (https://github.com/project-koku/koku-ui/blob/master/Dockerfile) when we added the PF3 ListView components. I would confirm that the cost details page still works and you may look at updating the Dockerfile as it would be good to get back to building as production there as that will make the app much smaller.

boaz1337: Yes, the `Cost Details` page doesn't render and I see this error message in the console:

```
TypeError: Cannot read property 'displayName' of undefined
    at Toolbar.js:18
    at Array.filter (<anonymous>)
    at h (Toolbar.js:17)
    at Fa (react-dom.production.min.js:173)
    at pi (react-dom.production.min.js:202)
    at hi (react-dom.production.min.js:203)
    at Yi (react-dom.production.min.js:212)
    at Ki (react-dom.production.min.js:211)
    at $i (react-dom.production.min.js:211)
    at Vi (react-dom.production.min.js:209)
```

boaz1337: Update: when using `production` in webpack `patternfly-react` stops working:
 - components like `Filter.TypeSelector` are undefined
 - `Toolbar` fails to render

To fix the first problem, call to `FilterTypeSelector`, `ListViewItem` and etc instead of `Filter.TypeSelector`, ListView.Item` and etc respectively.

I still didn't troubleshoot the second problem as I have no way to investigate it.
Trying to change the code in `node_module` seems to make any change,

I'm not sure whether it's related to our `webpack.config.js` or a problem on `Patternfly` side.

chambridge: @dlabrecq  You think its worth reaching out to @dmiller9911 on this ^ ?

He may have some thoughts since he did the webpack setup.

dmiller9911: My guess for this would be a PatternFly side with how it exports those subcomponents by mutating the module in the index.js. there is a `"sideEffects": false` false param in package.json would you be able to try and change that to `true` in `node_modules/patternfly-react/package.json` and retry. It is possible that could break the tree shaking. Webpack may think that it is not used and remove it. If that works patternfly-react will need that set to true which will sadly remove all of tree shaking for it causing the whole package to be imported even if it is not all used. The other option for it would be to stop doing that. Rollup did not support it at all so it would be a good change, just a big breaking one. Most consumers of `patternfly-react` at using `< v4 webpack` which does not support the sideEffects flag. So it is likely that many do not see the issue.

dlabrecq: I tried setting `"sideEffects": true` with PF3, PF-NEXT. For kicks, I also tried react-core, react-icons, react-styles, & react-tokens with no effect.

When I turn off hot deploy, I can at least see the dashboard and the following errors on the cost details page.

screen shot 2018-09-26 at 6 55 42 pm

boaz1337: I haven't worked on it much today but I think I have a way to troubleshoot it.
I will try working on it tomorrow. Nevertheless, if it takes too long, it's a good idea to postpone it till we move `detailsToolbar` to PF-next.

WDYT?

dlabrecq:

Built with commit a0fedd9bdc30a2a0abc4e076de7d9ac704ee9f55

https://deploy-preview-135--project-koku.netlify.com

package.json
boaz1337: OK - I found the source of the problem.
As mentioned by @dmiller9911 it's related to the subcomponents.

The Toolbar is using `Grid.Row` and `Grid.Col` which are in production `undefined` causing the entire page to fail and seeing those two error messages.

boaz1337: BTW - changing the `sideEffect` flag to `true` fixed that problem.

boaz1337:
To conclude - there is an issue with running patternfly-3 in production which requires to change `sideEffect` to `true` in the `package.json` file. Or alternatively not to use `subcomponents` in the code or any patternfly-3 component that is using `subcomponent` inside of it.

It seems to me that this should be fixed in other PR(s):
 - moving from patternfly-3 to patternfly-4 in details toolbar
 - asking patternfly-3 to change `sideEffect` to `true`
 - change `Toolbar` in patternfly-3 to use `Col` and `Row` instead of `Grid.Col` and `Grid.Row` respectively

It feels like the first option is the right way to go.

At any rate, the main point of this commit was around the fact that test packages like `jest` and `enzyme` were required in order to run the UI (both on production/development).

dlabrecq:
@boaz1337 Can you please file an issue with https://github.com/patternfly/patternfly-react?

dlabrecq:
@boaz1337 This PR has been merged https://github.com/project-koku/koku-ui/pull/141. Can you please retest to see if there are any further conflicts?

boaz1337:
Rebased and the cost details is loading and working. Thanks @dlabrecq.
In addition, no new errors/warning were logged in the console and no issues appeared in the UI.

PR #137, opened 2018-09-22
Add sort functionality

dlabrecq:
The sort dropdown and direction indicator are now functional. The query params update properly and state can be restored upon a page refresh, allowing bookmarking. Although, the sort has been intentionally disabled for now.

The "Cost Details" and "All Services" links have been fixed to also include the query params as expected. Previously, clicking on the "Cost Details" link (a second time) would lose all query params.

Note that order_by has been temporarily omitted from the request. This is due to the issue below when order_by is included, generating bad requests. When this issue has been resolved, we can easily enable the sort in another PR.

https://github.com/project-koku/koku/issues/375

### DIFF
--- a/src/api/query.ts
+++ b/src/api/query.ts
@@ -16,9 +16,17 @@ interface GroupBys {
   region?: GroupByValue;
 }
 
+interface OrderBys {
+  account?: string;
+  region?: string;
+  service?: string;
+  total?: string;
+}
+
 export interface Query {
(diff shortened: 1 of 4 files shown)


dlabrecq:

Built with commit 6191430777b67a625c7c9527d3d569bc1c1fc4b0

https://deploy-preview-137--project-koku.netlify.com

PR #140, opened 2018-09-26
fix(detailsToolbar): show filter type selector

boaz1337:
Filter.TypeSelector is not rendered because it has only one type right now which is confusing.

This is a small fix - adding a placeholder will display the type selector.

## Before
screenshot_2018-09-26 koku ui 3

### default
screenshot_2018-09-26 koku ui

### choosing filter type
screenshot_2018-09-26 koku ui 1

### selecting the placeholder
screenshot_2018-09-26 koku ui 2

### DIFF
--- a/src/pages/costDetails/detailsToolbar.tsx
+++ b/src/pages/costDetails/detailsToolbar.tsx
@@ -250,6 +250,7 @@ export class DetailsToolbar extends React.Component<DetailsToolbarProps> {
       <Toolbar>
         <Filter>
           <Filter.TypeSelector
+            placeholder={'Select Filter Type'}
             filterTypes={this.props.filterFields}
             currentFilterType={currentFilterType}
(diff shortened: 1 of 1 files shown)


dlabrecq:

Built with commit b8706c3d18d09a7631f92159b69ca4ea13ab8ed3

https://deploy-preview-140--project-koku.netlify.com

dlabrecq:
Note that when you choose the "Select filter type" option, we end up with this.

screen shot 2018-09-26 at 9 48 55 am

boaz1337:
You're right. I am going to close this now.
However, I think patternfly should have rendered a dropdown because it's not obvious what type we are using.

dlabrecq:
I thought it was odd, too, but why have a selection if you cannot make any choices. The input placeholder already describes what to do.

We'll replace this with PF4 shortly.

PR #141, opened 2018-09-28
Update pf-react to support webpack production

dlabrecq:
This pf-react update allows webpack 4 to run in production mode. Tested with `container:test` and `webpack-dev-server --env production`

Note that there may still be issues that pf-react must address. However, this update appears to be working with the PF3 components currently in use on the cost details page.

See: https://github.com/patternfly/patternfly-react/pull/564

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ RUN printf "[nginx]\nname=nginx repo\nbaseurl=http://nginx.org/packages/centos/7
 
 RUN chmod 777 /var/log/nginx && chmod 777 /var/cache/nginx && chmod 777 /var/run && rm -rf /var/log/nginx/* && rm -rf /var/cache/nginx/*
 
-ENV NODE_ENV=development \
+ENV NODE_ENV=production \
     NODE_PORT=8080
 
 WORKDIR /opt/koku-ui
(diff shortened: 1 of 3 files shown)


dlabrecq:

Built with commit 150b778d915b502f7b71dbf3f90e4bada848a0a6

https://deploy-preview-141--project-koku.netlify.com

chambridge:
So the only disconnect I have is that the container still has a development environment:
https://github.com/project-koku/koku-ui/blob/master/Dockerfile#L12

And I don't see how webpack comes into play in the end as we just seem to call `build`, maybe that is a mistake?
https://github.com/project-koku/koku-ui/blob/master/Dockerfile#L20

dlabrecq:
In our Docker file, you'll also see:

```
RUN yarn --production --non-interactive --silent \
    && yarn build >/dev/null
```

I believe that's creating a webpack production build via the following commands in package.json.

```
"build": "yarn build:client",
    "build:client": "yarn clean && webpack --env production",
```

Not quite sure how this is used in our Docker file?

`NODE_ENV=development`

boaz1337:
Yep - fixed the issue with the `cost details` page.

PR #149, opened 2018-10-03
fix(package.json): tslint hack to check test files too

boaz1337:
`tsconfig.json` is used by both `webpack` and `tslint`. `tslint` is using it to determine what files to run lint checking.

If a file is passed to `tslint` which is excluded `tslint` prints an error message saying the file was excluded and quits.
    
In `tsconfig.json` test files were excluded because they are not necessary for the UI. However, they are necessary for linting.
    
To fix this, I am using two different ts-configs. `tsconfig.lint.json` for tslint and `tsconfig.json` for webpack.

That way test files can be linted successfully and not be necessary to run the UI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "clean": "rimraf public/",
     "lint": "yarn lint:ts && yarn lint:locales",
     "lint:locales": "node scripts/syncLocales --check",
-    "lint:ts": "tslint --project tsconfig.json",
+    "lint:ts": "tslint --project tsconfig.lint.json",
     "precommit": "lint-staged",
     "start": "webpack-dev-server",
     "start:dev": "APP_NAMESPACE=koku-dev yarn start",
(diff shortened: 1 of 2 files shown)


dlabrecq:

Built with commit d6a6e88e371e71a805edb83ad6597c71cda32e3e

https://deploy-preview-149--project-koku.netlify.com

boaz1337:
Another way to solve this is to add to package.json under "scripts":
```json
"lint:precommit": "tslint",
```

and update "lint-staged" to
```json
    "*.{js,jsx,ts,tsx}": [
      "yarn lint:precommit --fix",
      "prettier --write",
      "git add"
    ]
```

This is good enough too because `husky`/`lint-staged` is passing to `yarn lint:precommit --fix` the files that were changed and thus there is no point in using `tsconfig.json`.

PR #150, opened 2018-10-04
feat(pages): add provider type selector in providersModal

boaz1337:


chambridge:
There is an API to call, that was show during the demo for last sprint.

**POST** `/api/v1/providers/`
Example Body:
```
{
    "name": "myocpcluster",
    "type": "OCP",
    "authentication": {
        "provider_resource_name": "my-ocp-cluster-id"
    }
}
```

OCP provider does not need the Bucket Name portion of the modal.

The Provider resource name is going to be a cluster identifier for now.

Also in the current AWS place holders could we make them generic instead of the content that is currently there please.

boaz1337:
OK - I rebased + added unit tests. Please give another look to make sure I didn't miss anything.
Thanks.

// @dlabrecq @chambridge

PR #179, opened 2018-10-16
fix(formGroup): Do not override children id props

### LINKED ISSUE #113: Form Group overrides test IDs
The FormGroup component unexpectedly overrides our test IDs used by QE. If we continue using 'testIds', we should ensure that isn't overridden.

As an example, consider the text inputs for the login screen. For the username input, I expect to see id='username-input' in the rendered HTML. Instead, I see id='Username0', which is the label name given to the FormGroup concatenated with a unique index.

```
<FormGroup label={t('login.username_label')}>
  <TextInput
    {...getTestProps(testIds.login.username_input)}
    ...
  />
</FormGroup>
```

boaz1337:
Fixes #113 

Although I couldn't reproduce what @dlabrecq wrote. I did find that the `id` prop is being overridden by RandomId children function.

This commit fixes this problem by setting the `randomId` as `data-label-for` instead of `id`.
Hope this is good enough.

In addition, added unit test to make sure it won't happen again.

// cc @dlabrecq @LandoCalrizzian

### DIFF
--- a/src/components/formGroup/formGroup.test.tsx
+++ b/src/components/formGroup/formGroup.test.tsx
@@ -1,0 +1,21 @@
+import { mount } from 'enzyme';
+import React from 'react';
+import { FormGroup, formGroupIdProp } from './formGroup';
+
+test('render form group', () => {
+  const labelName = 'username';
+  const idProp = 'div-id';
+  const testProp = 'div-test-id';
+  const view = mount(
+    <FormGroup label={labelName}>
+      <div id={idProp} testid={testProp} />
+    </FormGroup>
+  );
+  expect(view.find('label').text()).toBe(labelName);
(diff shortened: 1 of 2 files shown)


dlabrecq:

Built with commit edb58b2e67288a548fbb237ae77bbe1f76ab2c8c

https://deploy-preview-179--project-koku.netlify.com

dlabrecq:
The best example I can provide is the login page -- you can still test this via localhost:8002.

```
<FormGroup label={t('login.username_label')}>
  <TextInput
    {...getTestProps(testIds.login.username_input)}
   ...
  />
</FormGroup>
```

This outputs a random ID instead of using the ID provided to the text input.

screen shot 2018-10-16 at 9 39 57 am

I was thinking that we could provide the same ID to FormGroup via a prop?

PR #215, opened 2018-10-22
[WIP] refactor(details): break aws/ocpDetails into small components

boaz1337:
This is WIP. It is just hard to read a code with more than 100 LOC.

I am not sure if I am going to test components using snapshot, although it makes the testing really fast. I will probably figure out a better way to test stateless components later.

TODO:
 - [x] Break the header into components
 - [ ] Break List Items into components
 - [ ] Move Logic to Redux/utils

### DIFF
--- a/src/components/list/GroupBySelector.tsx
+++ b/src/components/list/GroupBySelector.tsx
@@ -1,0 +1,53 @@
+import { Select, SelectOption, Title } from '@patternfly/react-core';
+import { css } from '@patternfly/react-styles';
+import React from 'react';
+import { noop } from '../../utils/noop';
+import { styles } from './header.styles';
+
+interface GroupByOption {
+  value: string;
+  label: string;
+}
+
+interface GroupByProps {
+  title: string;
+  label: string;
+  onChange: (value) => void;
+  options?: GroupByOption[];
+  initValue?: string;
+}
+
+export const GroupBySelector: React.SFC<GroupByProps> = ({
+  title,
+  label,
+  onChange,
+  options,
+  initValue,
+}) => (
+  <React.Fragment>
+    <Title size="2xl">{title}</Title>
+    <div className={css(styles.groupBySelector)}>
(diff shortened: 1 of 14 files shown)


src/components/list/ListHeader.tsx
```diff
@@ -0,0 +1,11 @@
...
+import React, { ReactNode } from 'react';
+import { styles } from './header.styles';
+
+interface ListHeaderProps {
+  children: ReactNode;
+}
+
+export const ListHeader: React.SFC<ListHeaderProps> = ({ children }) => (
+  <header className={css(styles.header)}>{children}</header>
+);
```
dlabrecq: I don't know that we should componitize the list as AWS and OCP may not have the exact layout in the future? 

Considering the PF4 data list component should be ready after the next PF4 sprint, perhaps we should wait to update that, first?

src/components/list/GroupBySelector.tsx
```diff
@@ -0,0 +1,53 @@
...
+      onChange={onChange}
+    >
+      {options.map(option => (
+        <SelectOption
+          key={`group-by-${option.value}`}
+          value={option.value}
+          label={option.label}
+        />
+      ))}
+    </Select>
```
dlabrecq: Do we need a GroupBySelector component to wrap a PF4 select component?

src/components/list/Toolbar.tsx
```diff
@@ -0,0 +1,54 @@
...
+      </ToolbarItem>
+    </ToolbarGroup>
+    <ToolbarGroup>
+      <ToolbarItem>
+        <Button variant="plain" aria-label="download as CSV">
+          <DownloadIcon /> Export
+        </Button>
+      </ToolbarItem>
+    </ToolbarGroup>
+  </PFToolbar>
```
dlabrecq: Is this the PF4 toolbar? Where is the filter and sort? I don't believe we should add this toolbar until that functionality is available.

PR #257, opened 2018-11-06
Remove insights:proxy target

dlabrecq:
Since the run.sh script must be executed within insights-proxy repo, this target is no longer valid -- it needs an absolute path to LOCAL_CHROME.


dlabrecq:

Built with commit 062cae3dada5fd45b615369e33f71c2b12928871

https://deploy-preview-257--project-koku.netlify.com

chambridge:
you shouldn't have to run it with local chrome. 

The `run.sh` script just creates the command like the above that you are removing.

I'm okay with this being removed, just suggesting that it should still work.

dlabrecq:
You're correct. 

I saw that the run.sh script is adding `-v /Users/dlabrecq/Work/hccm/repos/insights-proxy:/chrome` and thought it was required.

Issue #184, opened 2018-10-16
Replace select element with PF4


dlabrecq:
## User Story
Replace the "Cost By" select element, in the AWS & OCP detail pages, with the PatternFly 4 select component.

## Impacts
UI

## Assumptions
The PatternFly 4 component can achieve similar visualizations.

## UI Details

The React based select component is now available for PatternFly 4

http://patternfly-react.surge.sh/patternfly-4/components/select

## Acceptance Criteria

The Cost Details page provides the same level of functionality with the new PF4 components

dlabrecq:
I believe we should to make this a dropdown instead.

Refer to https://github.com/project-koku/koku-ui/issues/235

PR #259, opened 2018-11-06
Styling Select to look like Dropdown

### LINKED ISSUE #235: Group by dropdown looks inconsistent with other dropdown component
The group by dropdown component found in both Cloud Cost and OpenShift Charge pages, is a PF3 component, but until we get a PF4 component to replace it, we would like the current component styled as if it is PF4 (to be more in-line with what the Sort by dropdown looks like)

1. Go to Cloud Cost or OpenShift Charge page
2. On the header of the page, underneath the title is line that allows the user to group the contents of the page by the various options in the dropdown.

openshiftchargesgroupbr

Would like the drop down line (label and control) to be akin to whatever we are doing for PF4, which I heard is closely like the style of the SortBy dropdown on the page.

christiemolloy:
This PR closes #235. 

The goal was to style the select component like the dropdowns in PF4. 

I tried to do as much as I could, but I had trouble overriding the select arrow and replacing it with the icon we use in pf4, as well as restyling options. I think this is as much as I can do unless we change the actual component to be a dropdown rather than a select.

### DIFF
--- a/src/pages/awsDetails/awsDetails.styles.ts
+++ b/src/pages/awsDetails/awsDetails.styles.ts
@@ -2,6 +2,7 @@ import { StyleSheet } from '@patternfly/react-styles';
 import {
   global_BackgroundColor_100,
   global_BackgroundColor_300,
+  global_BorderColor_light,
   global_BorderRadius_sm,
   global_BoxShadow_sm,
   global_Color_100,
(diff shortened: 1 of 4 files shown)


dlabrecq:
We have a PR to replace the select with a PF4 component. Instead of trying to style the select to look like PF4, it would be best to simply swap it for the PF4 component.

https://github.com/project-koku/koku-ui/issues/184

christiemolloy:
@dlabrecq agreed! I'm happy to have a go at replacing it but it might take me a while since I'm not a React expert. Is that something I should handle or you?

dlabrecq:
If you want to try it, go for it. Otherwise, I can pick this one up.

dlabrecq:
Closing in favor of https://github.com/project-koku/koku-ui/issues/184

Issue #242, opened 2018-10-26
Overview - All Cards: spacing between the title of the card and the hero value seems too tall


nlcwong:
Visually the "hero value" of the card seems to be floating on its own. 

1. Go to Cost Management > Overview

![spacebtwntitlevalue](https://user-images.githubusercontent.com/31672086/47588542-83f1f880-d934-11e8-917b-4498761acb39.png)

Would like it to be closer to the title of the card.

PR #262, opened 2018-11-07
Remove padding to move hero value closer to title

christiemolloy:
This closes issue #242 

Padding removed from card header and body so that hero value is moved closer to title.

### DIFF
--- a/src/components/reportSummary/reportSummary.styles.ts
+++ b/src/components/reportSummary/reportSummary.styles.ts
@@ -5,14 +5,9 @@ export const styles = StyleSheet.create({
   reportSummary: {
     height: '100%',
   },
-  cardHeader: {
-    paddingBottom: '0px !important',
-  },
-  cardBody: {
-    paddingTop: '0px !important',
-  },
   subtitle: {
     fontSize: global_FontSize_xs.value,
     color: global_Color_200.var,
+    marginBottom: '0',
   },
 });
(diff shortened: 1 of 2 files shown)


dlabrecq:
Can you please include a screenshot?

christiemolloy:
screen shot 2018-11-07 at 1 08 11 pm

src/components/reportSummary/reportSummary.styles.ts
```diff
@@ -5,6 +5,12 @@ export const styles = StyleSheet.create({
   reportSummary: {
     height: '100%',
   },
+  cardHeader: {
+    paddingBottom: '0px !important',
+  },
+  cardBody: {
+    paddingTop: '0px !important',
```
dlabrecq: Are we overriding the padding of PF cards? If so, perhaps we should confirm with Natalie we really want to do this?

Also, is there a way to do this without `!important` (e.g., perhaps by adding some specificity)? There are some examples of how to override PF styles in files like `awsDetails.styles.ts`.

christiemolloy: @dlabrecq You're right that wasnt the best approach. Instead of that, I've taken off margin under the botttom of the paragraph tag. Thats as much as I can do without overriding padding of cards

christiemolloy: screen shot 2018-11-07 at 1 57 25 pm

src/components/reportSummary/reportSummary.styles.ts
```diff
@@ -8,5 +8,6 @@ export const styles = StyleSheet.create({
   subtitle: {
     fontSize: global_FontSize_xs.value,
     color: global_Color_200.var,
+    marginBottom: '0',
```
dlabrecq: I like this approach better. Thanks

Issue #274, opened 2018-11-12
Use utilities instead of custom CSS on overview page


mcoker:
**Describe the bug**
The overview component uses custom CSS to separate the page title from the button to the right. We can use modifier classes for that instead.

PR #275, opened 2018-11-12
chore(overview): replace custom CSS with utility classes

mcoker:
fixes https://github.com/project-koku/koku-ui/issues/274

@dlabrecq we're pulling in fairly large utility class stylesheets to make this change. While I think it's beneficial to use the utilities and write as little custom CSS as possible, do you think it's worth it in this case to replace the custom CSS here?

### DIFF
--- a/src/pages/overview/overview.tsx
+++ b/src/pages/overview/overview.tsx
@@ -5,7 +5,6 @@ import {
   Title,
   TitleSize,
 } from '@patternfly/react-core';
-import { css } from '@patternfly/react-styles';
 import { Providers } from 'api/providers';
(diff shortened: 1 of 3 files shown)


src/pages/overview/overview.tsx
```diff
@@ -18,7 +17,6 @@ import { uiActions } from 'store/ui';
 import { getTestProps, testIds } from 'testIds';
 import AwsDashboard from '../awsDashboard';
 import OcpDashboard from '../ocpDashboard';
-import { styles } from './overview.styles';
```
dlabrecq: Can we also delete overview.styles.ts?

Issue #250, opened 2018-11-02
Overview page requires multiple refresh attempts before provider data is displayed


LandoCalrizzian:
**Describe the bug**
When accessing the overview page after first login, the "Add Provider" page is displayed indicating there is no provider data even though one or more providers returned by the api. The page requires 2-3 refresh attempts before the provider data is displayed in the UI

**To Reproduce**
1. Login to Insights and access Cost Management
2. Overview page only displays text for adding a provider because no provider was found
3. Refresh the page 2-3 times to populate the UI w/ provider data

**Expected behavior**
Provider data is queried and displayed when providers are present

**Desktop (please complete the following information):**
 - Browser Chrome 69 and Firefox 62

**Additional context**
Once data is cached in the session there are no issues

dlabrecq:
This is an issue with the Insights platform related to tokens

chambridge:
Insights team has provided a promise that we can wrap our API calls in to make sure we always have a token:
https://docs.google.com/document/d/1xjITNYbp1aw9Xu3lYLQppKQl-Rtj6_BaThwy1Zkazm4/edit#heading=h.8aucqk7x2oqb

We need to implement this in our code.

chargio:
+1

This is still happening @dlabrecq I can show you if you need to see it

vestival:
it happens the same to me

chargio:
![image](https://user-images.githubusercontent.com/7500590/49285287-55b58a80-f497-11e8-860e-c77f8d8b13a6.png)

dlabrecq:
I've opened a new Insights issue. 
https://github.com/RedHatInsights/insights-chrome/issues/76

lcouzens:
This is not completely blocking UI automation but is holding it up a lot, is there anything we can do to resolve it sooner? I will raise priority to high.

chambridge:
I messaged into frontend channel on slack to try and get some eyes on the issue.

boaz1337:
As mentioned in https://github.com/RedHatInsights/insights-chrome/issues/76
I believe the cause of this is the timeouts between the frontend and the backend.
We can solve this in two ways:

 1) Stretch the timeout for 7 seconds. This is good for short-term as this might fix the problem for 90% but what happens if the latency will be higher. In addition, this is not really user friendly to wait longer.

[...]

PR #281, opened 2018-11-15
Invoke Insights user promise

dlabrecq:
Invoking the Insights user promise before making axios calls. This appears to resolve the issue where the browser must be refreshed upon first login.

Fixes https://github.com/project-koku/koku-ui/issues/250

See https://docs.google.com/document/d/1xjITNYbp1aw9Xu3lYLQppKQl-Rtj6_BaThwy1Zkazm4/edit#

### DIFF
--- a/src/api/providers.ts
+++ b/src/api/providers.ts
@@ -49,10 +49,34 @@ export interface Providers {
 
 // See: http://koku-koku-dev.1b13.insights.openshiftapps.com/apidoc/index.html#api-Provider-createProvider
 export function addProvider(request: ProviderRequest) {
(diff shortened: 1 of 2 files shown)


src/api/users.ts
```diff
@@ -11,5 +11,7 @@ export function getCurrentUser() {
 }
 
 export function getUser(userId: string) {
```
chambridge: I don't think this gets called any more.

dlabrecq: Ok, I reverted that change. I'll file an issue to clean up the unused user calls.

PR #286, opened 2018-11-19
Update delta param to string with defaults

adberglund:
## Summary
I _believe_ this will accommodate the change for delta from being a Boolean to a string.

* The accepted delta value for AWS is currently `total`. 
* The accepted delta values for OCP are currently `charge`, `usage`, and `request` for memory and cpu inventory APIs, and `charge` for the charge API. The default value is `charge`.

### DIFF
--- a/src/api/query.ts
+++ b/src/api/query.ts
@@ -24,7 +24,7 @@ interface OrderBys {
 }
 
 export interface Query {
-  delta?: boolean;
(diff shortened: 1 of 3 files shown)


chambridge:
Can we update the example in swagger?
https://github.com/project-koku/koku/blob/master/docs/source/specs/openapi.yml#L1152

Issue #311, opened 2018-11-28
Refactor for OCP


dlabrecq:
The existing dashboard and detail pages are heavily typed for AWS, the code should be refactored so OCP can have its own properties, queries, axios paths, etc. In addition, the OCP dashboard will have a different layout, with new charts & navigation per the latest mocks.

PR #312, opened 2018-11-28
Refactor AWS and OCP reports

dlabrecq:
Since the existing dashboard and details functionality is heavily typed for AWS, the code is being refactored a bit so OCP can have its own typed properties, queries, axios paths, etc. In addition, the OCP dashboard will have a different layout, with new charts & navigation per the latest mocks.

Note that the AWS APIs are still in use; however, these changes will allow us to introduce OCP functionality without affecting existing AWS pages.

Fixes https://github.com/project-koku/koku-ui/issues/311


chambridge:
Unrelated but found this bug while running this branch:

image

dlabrecq:
Fixed the bad localization key. Thanks!

Issue #335, opened 2018-12-05
Cpu & memory cards to show last 30 days


dlabrecq:
**Describe the bug**
The cpu & memory cards for the OCP dashboard should show last 30 days. Currently, they display data for the last month.

**To Reproduce**
1. Go to Ocp dashboard
2. View CPU & memory usage cards

**Expected behavior**
Instead of showing data for the last month, show the last 30 days.

dlabrecq:
After speaking with @nlcwong & @chambridge, we decided not to show the last 30 days, but continue to show the current month. This will ensure the chart represents the current charge.

PR #336, opened 2018-12-05
Cpu & memory cards to show last 30 days

dlabrecq:
Per a review with @nlcwong , the cpu & memory cards for the OCP dashboard should show last 30 days. Currently, they display data only for the last month.

Fixes https://github.com/project-koku/koku-ui/issues/335

Screenshot
screen shot 2018-12-04 at 10 44 26 pm


dlabrecq:
After speaking with @nlcwong & @chambridge, we decided not to show the last 30 days, but continue to show the current month. This will ensure the chart represents the current charge.

Issue #337, opened 2018-12-05
OCP details page TypeError


dlabrecq:
**Describe the bug**
Due to some refactoring for the OCP dashboard page, wehad to change the URLs also used by the OCP details page. As a result, an error is generated because there is missing data -- the page is currently a copy of AWS and expects AWS data.

**To Reproduce**
1. Go to 'OpenShift Charge'

**Expected behavior**
Should not see a blank page. 

**Screenshots**
Uncaught (in promise) TypeError: Cannot read property 'isNumeric' of undefined
    at new z (detailsToolbar.tsx:16)
    at Ua (react-dom.production.min.js:174)
    at pi (react-dom.production.min.js:202)
    at hi (react-dom.production.min.js:203)
    at qi (react-dom.production.min.js:212)
    at Ki (react-dom.production.min.js:211)
    at $i (react-dom.production.min.js:211)
    at Vi (react-dom.production.min.js:209)
    at gi (react-dom.production.min.js:207)
    at Object.enqueueSetState (react-dom.production.min.js:148)

PR #338, opened 2018-12-05
Fix TypeError in OCP details page

dlabrecq:
Fixed a TypeError issue in OCP details page.

Note that the OCP page still needs some refactoring to use the OCP APIs properly. For example, the toolbar is missing the filter and sort. That will be updated via another PR. This just gets the page working after the work to refactor the OCP dashboard.

Fixes https://github.com/project-koku/koku-ui/issues/337

Sreenshot
screen shot 2018-12-05 at 10 40 09 am

### DIFF
--- a/src/locales/es.json
+++ b/src/locales/es.json
@@ -165,7 +165,7 @@
       "name": "Sort by: Name"
     },
     "title": "OpenShift Charges",
-    "total_charge": "Total OpenShift Charge"
+    "total_charge": "Total Charge"
   },
   "overview": {
     "aws": "Amazon Web Services",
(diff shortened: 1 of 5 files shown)


src/locales/es.json
```diff
@@ -47,7 +47,7 @@
       "name": "Sort by: Name"
     },
     "title": "Cloud (AWS) Cost",
-    "total_cost": "Total Cost"
+    "total_cost": "Total Charge"
```
chambridge: Seems like the wrong update.

dlabrecq: Yes, I changed the wrong one. Thanks!

Issue #363, opened 2018-12-16
Add limit to bullet chart legend


dlabrecq:
**Describe the bug**
Add limit to bullet chart legend in OCP details page.

Currently, we can show the limit as the threshold (i.e., a vertical red line), but there's no tooltip or any other indication of what the value represents. Ideally, the PF3 component would have a flag so we can include the limit (threshold) in the legend. Then, we could add an appropriate label.

We can create our own custom legend, but prefer to use the default PF component behavior. When creating a custom legend, we must also provide our own custom colors for both the legend and values in the chart, in addition to providing custom tooltips.

Note there is a PF3 issue to add the limit (threshold) to the legend. See: https://github.com/patternfly/patternfly-react/issues/1061

**To Reproduce**
1. Go to OCP details page
2. Click on any toggle to expand a list row

**Expected behavior**
We want to show the limit (threshold) as part of the legend.

**Screenshots**
![screen shot 2018-12-15 at 10 10 37 pm](https://user-images.githubusercontent.com/17481322/50049615-00909000-00b7-11e9-82a2-767b7311f749.png)

PR #366, opened 2018-12-18
Add limit to bullet chart legend

dlabrecq:
Added the cpu and memory limits to the bullet chart legend. This should be the last round of PF3 fixes to add the limit (threshold) to the legend and remove our custom code.

Fixes https://github.com/project-koku/koku-ui/issues/363

screen shot 2018-12-18 at 3 40 54 pm


package.json
```diff
@@ -33,12 +33,12 @@
   "dependencies": {
     "@babel/core": "7.0.0-beta.47",
     "@fortawesome/free-solid-svg-icons": "^5.1.0",
-    "@patternfly/patternfly-next": "1.0.79",
-    "@patternfly/react-charts": "1.1.0",
-    "@patternfly/react-core": "1.36.0",
-    "@patternfly/react-icons": "2.8.0",
+    "@patternfly/patternfly-next": "1.0.106",
```
chambridge: Will we need Insights to bump any of their dependencies for this? If so, do we open an issue on that?

dlabrecq: I don't believe there are any breaking changes. I'm mainly interested in the bug fixes. I'll open an issue if we encounter any conflicts.

boaz1337: Theoretically, the changes in `1.0.106` should be backward compatible with `1.0.79` but it always a good idea to make sure it doesn't actually break the UI.

Anyway, LGTM. :+1:

PR #378, opened 2018-12-30
chore(nav,masthead,sidebar,backdrop,page): remove old components

boaz0:
It looks like `verticalNav`, `masthead`, `sidebar`. `backdrop` and `page` are coming from `insights` and therefore are no longer required.


boaz1337:
Hi @dlabrecq 
I agree - that's a good idea. I am just not sure whether adding a `deprecated` comment in files headers or to append `deprecated` next to the filenames: `sidebar.deprecated.js`.

I am leaning towards adding deprecated-comment in the file header, though.

What do you think?

dlabrecq:
I was originally thinking of moving components to another, "archive" directory so we wouldn't have to build them. Then components, like backgroundImage, wouldn't be subject to PF4 changes, for example.

I believe a comment in the header would be fine. That would be an easy way to see what is still in use.

boaz0:
closing this as we might be using those components in the future.

Issue #392, opened 2019-01-07
OCP details: Show project tags


dlabrecq:
**Describe the bug**
Show project tags on the OCP details page.

~~Note: Need input from @nlcwong as to what clicking on a tag will do. We may want to apply a filter?~~

See: https://redhat.invisionapp.com/share/ENPJQTBXTAC#/screens/336165164

**To Reproduce**
1. Go to 'OpenShift Charge'
2. See error

**Expected behavior**
- Applies to group by projects only
- Should see a list of tags per project
- ~~Show first 5 tags only?~~ Show all tags (for now)
- Clicking on a tag button applies a filter (for now)

**Screenshots**
![screen shot 2019-01-07 at 10 58 45 am](https://user-images.githubusercontent.com/17481322/50778152-3b778500-126b-11e9-9f9d-129d3ac5028c.png)

dlabrecq:
Here are some tag API examples (compliments of Chris) to get started:

https://api.access.ci.itop.redhat.com/r/insights/platform/cost-management/api/v1/tags/ocp/?filter[time_scope_units]=month&filter[time_scope_value]=-1&filter[resolution]=monthly

https://api.access.ci.itop.redhat.com/r/insights/platform/cost-management/api/v1/tags/ocp/?filter[time_scope_units]=month&filter[time_scope_value]=-1&filter[resolution]=monthly&key_only=True

https://api.access.ci.itop.redhat.com/r/insights/platform/cost-management/api/v1/tags/ocp/?filter[time_scope_units]=month&filter[time_scope_value]=-1&filter[resolution]=monthly&filter[project]=metering-koku

https://api.access.ci.itop.redhat.com/r/insights/platform/cost-management/api/v1/reports/inventory/ocp/cpu/?filter[time_scope_units]=month&filter[time_scope_value]=-1&filter[resolution]=monthly&group_by[tag:app]=prom

https://api.access.ci.itop.redhat.com/r/insights/platform/cost-management/api/v1/reports/charges/ocp/?filter[time_scope_units]=month&filter[time_scope_value]=-1&filter[resolution]=monthly&group_by[tag:app]=*

https://api.access.ci.itop.redhat.com/r/insights/platform/cost-management/api/v1/reports/charges/ocp/?filter[time_scope_units]=month&filter[time_scope_value]=-1&filter[resolution]=monthly&group_by[tag:app]=hive

PR #398, opened 2019-01-08
Add project tags to OCP details page

dlabrecq:
Adds tags to OCP details page for groupBy project. This also allows the user to filter by clicking on a tag.

Note that there are 20 tags for the "metering-koku" project. 

Fixes https://github.com/project-koku/koku-ui/issues/392

screen shot 2019-01-08 at 10 36 12 pm

### DIFF
--- a/src/api/ocpQuery.ts
+++ b/src/api/ocpQuery.ts
@@ -5,6 +5,7 @@ export interface OcpFilters {
   time_scope_units?: 'month' | 'day';
   resolution?: 'daily' | 'monthly';
   limit?: number;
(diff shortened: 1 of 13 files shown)


dlabrecq:
I better understand how to use the API now. However, I'm confused as to what tag key/values the user expects to see in the UI? For the "metering-koku" project, there are 20 values accross 6 keys. I'll attempt to clarify with @nlcwong.

dlabrecq:
Updated to use the tags API as intended. Just need to confirm we want to show all those tags?

Issue #407, opened 2019-01-16
Dropdown doesn't work with latest PF4 package


dlabrecq:
The Dropdown component no longer functions properly with PF4 v1.45.2. The details page is not receiving an onClick event. 

Need to revert to an earlier PF4 package version.

See PF4 issue:
https://github.com/patternfly/patternfly-react/issues/1195

PR #408, opened 2019-01-16
Fix dropdown onClick event

dlabrecq:
Update PF4 packages to fix the broken dropdown onClick event in the AWS/OCP detail pages.

Fixes https://github.com/project-koku/koku-ui/issues/407


boaz0:
LGTM :+1:

Issue #411, opened 2019-01-16
Cannot sort by clusters


dlabrecq:
Sorting by name does not work for clusters in the OCP detail pages. It appears that we're attempting to call the API with `order_by[cluster_alias]` as a parameter.

PR #412, opened 2019-01-16
Fix cluster sort in OCP details

dlabrecq:
Changed sort to use `cluster` instead of `cluster_alias` in OCP details.

Fixes https://github.com/project-koku/koku-ui/issues/411


boaz0:
LGTM :+1:

PR #425, opened 2019-01-22
Remove PF4 Dropdown onClick workaround

dlabrecq:
This replaces the `onClick` workaround added to the PF4 Dropdown item, which prevents the page from auto-submitting when a menu item is selected. The PF4 fix is to add `component='button'` to the DropdownItem component.

See issue: https://github.com/patternfly/patternfly-react/issues/1194


boaz0:
LGTM :+1:

Issue #424, opened 2019-01-21
Charts don't fit to the table


quarckster:
**Describe the bug**
Charts in "Cloud Cost" tab don't fit to the table.

**To Reproduce**
1. Go to "Cloud Cost"
2. Group cost by: Account
3. Expand some entry from the table

**Screenshots**
![screenshot_20190121_163114](https://user-images.githubusercontent.com/1011257/51483730-0fdfb900-1d9a-11e9-8ccf-fa0ee1c5399f.png)

PR #427, opened 2019-01-22
fix(pieChart): fix legend text cut

boaz0:
closes #424 

Now it looks like this:
ipd2

And on mobile:
ipd1

PR #428, opened 2019-01-22
Long pie chart labels are truncated

dlabrecq:
This allows long labels to overflow instead of being truncated.

Fixes https://github.com/project-koku/koku-ui/issues/424


dlabrecq:
Boaz also submitted PR https://github.com/project-koku/koku-ui/pull/427.

Issue #451, opened 2019-02-02
Add chart tic labels for x-axis


dlabrecq:
**Describe the bug**
As a user, I want to see tic labels on the chart x-axis. This will help indicate that we're looking at days of the month at a glance. 

**To Reproduce**
1. Go to 'Overview'
4. See charts

**Expected behavior**
Want to add tic labels to the AWS/OCP overview and OCP historical charts.

**Screenshots**
There are no mocks, but the x-axis should look similar to the bullet chart below.

<img width="511" alt="screen shot 2019-02-01 at 11 20 25 pm" src="https://user-images.githubusercontent.com/17481322/52159854-fc74fd80-2677-11e9-86e4-4e1db0b5d158.png">

**Additional context**
Only need to add the start (1), end (31), and a mid-month date (e.g., 15).

PR #454, opened 2019-02-04
Added tic labels to charts

dlabrecq:
Added a couple tic labels to the chart x-axis. This helps indicate the day of month at a glance (e.g., 1, 15, 31, etc.).

Fixes https://github.com/project-koku/koku-ui/issues/451

Historical charts:
screen shot 2019-02-04 at 1 49 56 pm

OCP charts:
screen shot 2019-02-04 at 1 50 36 pm

AWS charts:
screen shot 2019-02-04 at 1 50 27 pm

### DIFF
--- a/src/components/historicalTrendChart/historicalTrendChart.tsx
+++ b/src/components/historicalTrendChart/historicalTrendChart.tsx
@@ -13,9 +13,11 @@ import {
   getTooltipContent,
   getTooltipLabel,
 } from 'components/commonChart/chartUtils';
+import getDate from 'date-fns/get_date';
 import React from 'react';
 import { FormatOptions, ValueFormatter } from 'utils/formatValue';
 import { DomainTuple, VictoryAxis, VictoryStyleInterface } from 'victory';
+import { getDateRange } from '../commonChart/chartUtils';
 import { chartStyles, styles } from './historicalTrendChart.styles';
 
 interface HistoricalTrendChartProps {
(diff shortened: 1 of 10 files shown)


src/components/historicalTrendChart/historicalTrendChart.tsx
```diff
@@ -267,11 +267,15 @@ class HistoricalTrendChart extends React.Component<
               datum.cost.charts.map((chart, index) => {
                 return this.getChart(chart, index);
               })}
-            <VictoryAxis label={xAxisLabel} style={chartStyles.axis} />
+            <VictoryAxis
+              label={xAxisLabel}
+              style={chartStyles.xAxis}
+              tickValues={[1, 15, 31]}
```
chambridge: Does this not show `31` if the month doesn't have 31 days? Is the intent to show begin, middle, end, of which I feel like you could use the max of the two months (if you have two months of non-zero data).

src/components/historicalUsageChart/historicalUsageChart.tsx
```diff
@@ -458,11 +458,15 @@ class HistoricalUsageChart extends React.Component<
               datum.current.charts.map((chart, index) => {
                 return this.getChart(chart, index);
               })}
-            <VictoryAxis label={xAxisLabel} style={chartStyles.axis} />
+            <VictoryAxis
+              label={xAxisLabel}
+              style={chartStyles.xAxis}
+              tickValues={[1, 15, 31]}
```
chambridge: same comment as before about statically defining the ticks.

dlabrecq: My thought was if February has 28/29 days, and January was the previous month, the max would be 31. If February is the current month, the max is still 31.

    January - 31 days
    February - 28 days in a common year and 29 days in leap years
    March - 31 days
    April - 30 days
    May - 31 days
    June - 30 days
    July - 31 days
    August - 31 days
    September - 30 days
    October - 31 days
    November - 30 days
    December - 31 days

I suppose if we were to show only one month, then it should be dynamic.

dlabrecq: Ok, I've added code to make the end date dynamic. The default is 31, since that's the range (1-31 days) we intend to show even with zero data. However, if we only have one month to show (i.e., February), the end date could be shown as 28/29.

Issue #207, opened 2018-10-21
Replace PF3 listView with PF4 component


dlabrecq:
Note that this PF4 component is not currently available -- scheduled for 1.0.0-alpha.5 (Oct-Nov 9).
See https://docs.google.com/spreadsheets/d/17aEfvG1DZ0dZmpwacYaCzKv4sog0LuatXCwNizJ7HhE/edit#gid=205680820

## User Story
Replace the PF3 listView component with the PatternFly 4 data list component. This is currently being used in the AWS and OCP detail pages.

## Impacts
UI

## Assumptions
The PatternFly 4 component can achieve similar visualizations.

## UI Details
See http://patternfly-react.surge.sh/patternfly-4

## Acceptance Criteria

The pages achieve similar functionality with the new PF4 component.

karelhala:
I would like to point out that data list to show these data is probably wrong and Table should be used here. Data list are ment as alternative view to cards for instance because they lack certain actions and accessibility.

For instance Data list doesn't have any header to show at the top so no sorting is available as well. When table is shown on small devices it behaves somewhat differently - select boxes and actions are in different place.

So unless you are showing somewhat static data (less than 10 for each user) and we do not expect to filter, sort and paginate over these items consider using Table instead of Data lists.

dlabrecq:
This issue was more a placeholder to replace the PF3 list. I believe we will be more interested in using the table Vs data list in favor of features similar to the PF3 list -- both are available in pF4 now.

PR #466, opened 2019-02-08
OCP Details: Replace PF3 ListView with PF4 table

dlabrecq:
This replaces the old PF3 ListView with a new PF4 table.

Note that this is for the OCP details page only. AWS Details will be another PR.

Fixes https://github.com/project-koku/koku-ui/issues/207

Example:
screen shot 2019-02-08 at 2 02 51 am

Known issues with PF4 table: https://github.com/patternfly/patternfly-react/issues/1337


src/pages/ocpDetails/detailsTable.tsx
```diff
@@ -0,0 +1,327 @@
...
+          cells: [
+            item.label,
+            this.getMonthOverMonthCost(item, index),
+            this.getTotalCost(item, index),
+          ],
+          isOpen: false,
+          item,
+        },
+        {
+          parent: index,
```
karelhala: This is wrongly calculated parent. What you end up with is
```JS
[
 { cells: ['parent - 1'], isOpen: false }, // 0
 { cells: ['child - 1'], parent: 0}, // 1
 { cells: ['parent - 2'], isOpen: false }, // 2
 { cells: ['child - 2'], parent: 1 } // this is wrong, it should be number 2
]
```

To calculate parent correctly you need to use `index * 2`

src/locales/de.json
```diff
@@ -176,7 +176,6 @@
       "memory_usage": "Usage - {{value}} GB-Hours"
     },
     "change_column_title": "Month Over Month Change",
-    "charge_column_subtitle": " (Total {{total}})",
     "charge_column_title": "Charges",
```
chambridge: This line was changed in `en.json` but not here.

src/locales/es.json
```diff
@@ -176,7 +176,6 @@
       "memory_usage": "Usage - {{value}} GB-Hours"
     },
     "change_column_title": "Month Over Month Change",
-    "charge_column_subtitle": " (Total {{total}})",
     "charge_column_title": "Charges",
```
chambridge: This line was changed in `en.json` but not here.

src/pages/ocpDetails/detailsTable.tsx
dlabrecq: Ah, much better. Thank you!

src/locales/de.json
dlabrecq: Thanks. The linter didn't catch that.

src/locales/es.json
dlabrecq: Fixed. Thank you

PR #480, opened 2019-02-13
Updating to use a single esi tag in index.html

ryelo:
This change is needed after: https://github.com/RedHatInsights/insights-chrome/pull/150

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,16 +1,11 @@
 <!DOCTYPE html>
 <html lang="en">
-
-<head>
-  <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-  <meta name="theme-color" content="#000000">
-  <title>Cost Management</title>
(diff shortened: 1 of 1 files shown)


chambridge:
Don't merge this until after the 10-12 demo please.

ryelo:
@dlabrecq we haven't released the newest version of chrome to stable. This is going to work with the beta version of chrome.

dlabrecq:
I saw the Slack comment about "insights-chrome is updated on all 3 stable environments". The change appears to be working now.

Should we also update the title per the suggestion below? I can create another PR if you prefer?

```
<!DOCTYPE html>
<html lang="en-US">
    <head>
        <meta charset="UTF-8">
        <title>Red Hat Insights | Cost Management </title>
        <esi:include src="/@@insights/static/chrome/snippets/head.html"/>
    </head>
    <body>
        <esi:include src="/@@insights/static/chrome/snippets/body.html"/>
    </body>
</html>
```

ryelo:
@dlabrecq Ah. Yes. That would be great. As long as it's working, that's the biggest factor!

Issue #226, opened 2018-10-23
OCP Details page: no provider empty state


dlabrecq:
## User Story
As a user, I want to view an empty state when the OCP provider is not available.

See: https://redhat.invisionapp.com/share/ENPJQTBXTAC#/screens/336165151

## Impacts
UI

## Assumptions
- User has not configured an OCP provider

## UI Details
![screen shot 2019-01-07 at 10 02 51 am](https://user-images.githubusercontent.com/17481322/50775185-77a6e780-1263-11e9-80ba-32c8846b203c.png)

## Acceptance Criteria
The OCP details page has a similar layout as shown in the mocks.

PR #505, opened 2019-02-22
Add empty states for AWS/OCP Details

dlabrecq:
This adds some empty states for the AWS/OCP Details page, similar to the functionality previously added to the overview page.

- Show empty state when there are no providers
- Show no authorization state
- Show unexpected error state
- Show loading state

Since this is similar functionality, I've refactored the overview and located state components under the components directory for re-usability.

The "Group By" dropdown and other header items are hidden when empty / error states are shown. An empty state for the table itself will be added via another PR.

I've also refactored the providers request to support a query filter. This will be necessary to determine if there are any AWS and OCP providers, separately. Thus, avoiding multiple provider requests to parse paginated data. This feature will be enabled when https://github.com/project-koku/koku/issues/658 is available.

Fixes https://github.com/project-koku/koku-ui/issues/227
Fixes https://github.com/project-koku/koku-ui/issues/226

No Providers:
no providers

Loading:
sources

Unauthorized:
unauthorized

Unexpected Error:
unexpected


chambridge:
So I noticed yesterday when I triggered a 400 by inputting bad parameter (like a invalid bucket for the given ARN) the overview went to an Unexpected error state.

dlabrecq:
Yes, I saw that with Insights beta. I no longer see that after refactoring the code. I'm also only tracking the error state of the providers call.

You can see that the overview no longer displays an error in the background.
chrome-capture

Issue #535, opened 2019-02-26
Update Overview Layout


dlabrecq:
## User Story
As a user, I want to see a different layout for the overview page. That is, the total cost card should span the entire width of the page. Other card should appear below, stacked horizontally.

## Assumptions
- All tabs will use the same card layouts
-

## UI Details
See latest mock: https://redhat.invisionapp.com/share/MJQQ9LYYB7Q

Per my discussion with @nlcwong, we want to change the existing cards to support a horizontal layout. We also want the ability to hide the top services, accounts, and regions at the bottom of the card.

This work will help create layouts for the new OCP on AWS cards, shown via a new overview tab.

## Screenshots
![screen shot 2019-02-27 at 10 57 38 am](https://user-images.githubusercontent.com/17481322/53503737-a3ce3000-3a7e-11e9-83e4-c3df0c60fb81.png)

Issue #526, opened 2019-02-25
Move overview tabs to header


dlabrecq:
**Describe the bug**
Need to move the tabs to the header per Kyle's high fidelity mocks.

https://redhat.invisionapp.com/share/ENPJQTBXTAC#/screens/336165162

**To Reproduce**
1. Go to 'overview'
2. See error

**Expected behavior**
Tabs need to be moved from the page body to the page header.

**Screenshots**
Before:
![screen shot 2019-02-25 at 1 26 28 pm](https://user-images.githubusercontent.com/17481322/53359473-feda1880-3900-11e9-99eb-63b1b86e2249.png)

After:
![screen shot 2019-02-25 at 1 22 40 pm](https://user-images.githubusercontent.com/17481322/53359277-722f5a80-3900-11e9-9a02-20ae43c45452.png)

**Additional context**
This depends on refactoring the PF4 tabs component so it's tab children can be located outside the header.

See: https://github.com/patternfly/patternfly-react/issues/1364

PR #549, opened 2019-02-28
Overview layout update

dlabrecq:
This introduces a new layout for the overview page. For AWS, this design accommodates the two additional cards we're planning to add. Ultimately, this layout will also be used for the new OCP on AWS cards.

Latest mock: https://redhat.invisionapp.com/share/MJQQ9LYYB7Q

Fixes https://github.com/project-koku/koku-ui/issues/535

Screenshot:
screen shot 2019-02-27 at 6 02 42 pm


boaz0:
Hi @dlabrecq ,

Looking at the screenshot and the mocks, I see few nuances:
1) The location of the tabs - in the mock they on the same level of the title.
2) Charts progress bars - in the mock charts don't have progress bars while in the screenshot the progress bars are displayed under the charts.

At any rate, LGTM!

dlabrecq:
Thanks @boaz.

Regarding the overview tabs, it's not technically feaable to place the PF4 tabs in the header at this time. The page body would need to be a child of the tab, inside the header, which is not what we want. The PF4 team has an open PR to refactor the tabs component to make this possible.

https://github.com/project-koku/koku-ui/issues/526

Regarding the progress bars, we don't have a mock for the AWS/OCP tabs. Although, we want to use the same general layout for "AWS and OCP", while showing the progress bars here. This particular mock was created for the OCP on AWS tab -- in that case, we will have many cards and omit the progress bars.

Issue #551, opened 2019-02-28
Remove invalid test


dlabrecq:
In awsDashboardWidget.test.tsx and ocpDashboardWidget.test.tsx, we have the following test, which simulates a change event in our custom tabs component. However, this test no longer works since we replaced our custom tabs with PF4 tabs. It was suggested that this test should be omitted.

```
test('change tab triggers updateTab', () => {
  const tabId = AwsDashboardTab.regions;
  const view = shallow(<AwsDashboardWidgetBase {...props} />);
  view.find(Tabs).simulate('click', { event: {}, tabIndex: 2 });
  expect(props.updateTab).toBeCalledWith(props.id, { tabId });
});
```

"In order to simulate a click, we need to mount or fully render the component. Shallow is not good enough because when Tabs is rendered it mutates the given tabs so when you click on them onSelect is called. But shallow doesn't process that part. That's why the test failed."

"If we want to check that updateTab is triggered though we can simulate 'select' event but then we won't get the right arguments as select is not a valid event." -- @boaz0

PR #552, opened 2019-02-28
Invalid tabs test

dlabrecq:
This test had previously simulated a change event in our custom tabs component. However, the test no longer functions properly after replacing our custom tabs with PF4 tabs. Both @boaz0 and I attempted to refactor it, but we agreed the test should be removed for technical reasons.

For more details, please see: https://github.com/project-koku/koku-ui/issues/551


boaz0:
LGTM :+1:

PR #565, opened 2019-03-02
Fix json item to grab cost data from.

chambridge:


chambridge:
Feel free to revert, but this fixes data showing in *CI* so I'm putting it in so I can record the demo.

Issue #561, opened 2019-03-02
Empty state does not disappear after adding provider


lcouzens:
**Describe the bug**
Empty state does not disappear after adding provider

**To Reproduce**
1. Go to 'Overview' (without any providers)
2. Click on 'Add a source'
3. Add AWS or OCP source
4. Finish add source modal page stays on 'No Money, No Problem!'

**Expected behavior**
Add a source empty state should change to provider view.

**Screenshots**
![screenshot from 2019-03-02 14-54-50](https://user-images.githubusercontent.com/18352403/53683678-91145f00-3cfb-11e9-87fb-0d3e685ad188.png)

![screenshot from 2019-03-02 14-54-20](https://user-images.githubusercontent.com/18352403/53683676-8bb71480-3cfb-11e9-8151-d14390820e11.png)

**Desktop (please complete the following information):**
 - OS: Fedora
 - Browser Chrome
 - Version 66

**Additional context**
Work around Refresh browser after adding provider
Testing against CI insightsbeta

dlabrecq:
We likely need to fetch the providers list again, after the onboarding modal is complete. We only fetch when the page is updated -- opening/closing the onboarding modal does not update the page.

boaz0:
Dan, is it related to the onboarding?

boaz0:
Hi @dlabrecq put it on me. I think I know how to solve that.

PR #571, opened 2019-03-04
Refresh empty state after adding provider

dlabrecq:
This changes listens for the onboarding "Add Source" status and errors to see if the empty state should be refreshed.

Fixes https://github.com/project-koku/koku-ui/issues/561


dlabrecq:
FYI, I attempted to request providers via the store's addSource action, when addSourceSuccess is dispatched. That would be ideal; however, I could not trigger the empty state to update. By listening for onboardingStatus, each page can refresh itself when necessary. 

Will merge this, but if there is a better way, I'm all for it.

Issue #575, opened 2019-03-05
OCP on AWS date range


dlabrecq:
**Describe the bug**
OCP on AWS is not showing the month range properly.

**Expected behavior**
Expecting to see March 1st - March 5th

**Screenshots**

Total cost card:
![screen shot 2019-03-05 at 10 55 47 am](https://user-images.githubusercontent.com/17481322/53818086-87763b80-3f35-11e9-8344-3db7b0bcccf0.png)

AWS cards:
![screen shot 2019-03-05 at 10 55 53 am](https://user-images.githubusercontent.com/17481322/53818093-8a712c00-3f35-11e9-9c5a-83c70b59341e.png)

dlabrecq:
Just missed adding the localization string below.

`"widget_subtitle_plural": "$t(months.{{month}}) {{startDate}} - $t(months.{{month}}) {{endDate}}"`

PR #576, opened 2019-03-05
OCP on AWS date ranges

dlabrecq:
Missed adding a localization string.

Fixes https://github.com/project-koku/koku-ui/issues/575

Total cost:
screen shot 2019-03-05 at 11 01 44 am

AWS cards:
screen shot 2019-03-05 at 11 01 50 am


chambridge:
Missing es.json, and de.json updates.

dlabrecq:
Thanks. Wondering if we should just delete those extra files?

chambridge:
Your call on the locale files, I don't have a preference.

PR #577, opened 2019-03-05
Update Insights and PF4 packages

dlabrecq:
This just updates Insights and PF4 to the latest packages

Fixes https://github.com/project-koku/koku-ui/issues/578


boaz0:
Shouldn't we replace `Select` to `FormSelect` too?

dlabrecq:
@boaz0 Select was renamed in PF4 a while ago. Do you see another instance of Select anywhere?

boaz0:
Nope, I thought I was using Select in the onboarding page but I checked it and I am not.
LGTM :+1:

PR #580, opened 2019-03-06
UI change for time scope API updates.

chambridge:
Minimal impact UI update to have the new API which will return 2 months of data to "act" like old API.

Another PR can make performance improvements to alter API calls to use one API call and process the previous and current data from the single 2 months of data call.

Real impact here is on the historical charts previous/current data calls.
image

Associated with: https://github.com/project-koku/koku/pull/679


chambridge:
@dlabrecq this doesn’t optimized down to less calls. Can be done in a future PR as mentioned above. This strips out data to keep the current interaction.

Issue #606, opened 2019-03-11
Historical tool tips cut off be modal size


lcouzens:
**Describe the bug**
Historical tool tips cut off be modal size

**To Reproduce**
1. Go to 'details page'
2. Click on 'historical data'
3. Show first day tool tip

**Expected behavior**
Tool tip data should be readable

**Screenshots**
![Screenshot from 2019-03-11 11-30-33](https://user-images.githubusercontent.com/18352403/54121121-82136800-43f1-11e9-89c0-f943b738f7bb.png)
![Screenshot from 2019-03-11 11-29-45](https://user-images.githubusercontent.com/18352403/54121124-850e5880-43f1-11e9-8542-f1a2dff3cf96.png)

**Desktop (please complete the following information):**
 - OS: Fedora
 - Browser Firefox
 - Version 65

**Additional context**
Testing against CI insightsbeta

dchorvat1:
The modal also cuts off the bottom 'day of month' text when resized

dlabrecq:
The tooltip is provided by Victory, the underlying library for the charts. We could make the margin much larger, but that may be undesirable.

@dchorvat1 When resizing the window, the modal should be scrollable.

PR #607, opened 2019-03-11
Historical chart tooltips are cut off

dlabrecq:
This solves an issue where the historical chart tooltips are cut off for the first day or two. There just isn't enough room in the left side margin.

Ideally, we would just set the overflow to visible, which would display the tooltip in its entirety. However, this has the side effect of disabling vertical scrolling. In order to scroll vertically when the window is resized, the chart container cannot overflow. 

Instead, I've modified our PF4 Victory charts to use voronoiDimension='x', which forces the tooltip to be displayed within the chart's x-axis. This also stacks all tooltips for any given point on the x-axis, which is something we planned to do anyway.

Fixes https://github.com/project-koku/koku-ui/issues/606

Example:
Screen Shot 2019-03-11 at 10 35 00 AM


boaz0:
:+1:  if test is :green_heart:

Issue #597, opened 2019-03-08
AWS database services card


dlabrecq:
**Describe the bug**
Add network services card to AWS overview page.

**Expected behavior**
The new card should appear similar to the mock below.

Mock: https://redhat.invisionapp.com/share/HEO4ISHAT8Z#/screens/350009551

**Screenshots**
<img width="1178" alt="Screen Shot 2019-03-08 at 3 22 26 PM" src="https://user-images.githubusercontent.com/17481322/54053665-ffeb2f80-41b5-11e9-932b-81689072ee64.png">

Issue #596, opened 2019-03-08
AWS network services card


dlabrecq:
**Describe the bug**
Add network services card to AWS overview page.

**Expected behavior**
The new card should appear similar to the mock below.

Mock: https://redhat.invisionapp.com/share/HEO4ISHAT8Z#/screens/350009551

**Screenshots**
<img width="1382" alt="Screen Shot 2019-03-08 at 3 16 14 PM" src="https://user-images.githubusercontent.com/17481322/54053418-2d83a900-41b5-11e9-8032-f9ca09cbaeab.png">

PR #609, opened 2019-03-12
AWS database & network cards

dlabrecq:
This adds the AWS database & network services cards. And some refactoring to allow filtering to be applied generically across all overview tabs.

Note: The network card uses "Load Balancer-Network" as the product family for now.

@chargio will see if the price list identifies networking as something different. If so, I'll update the product family via another PR -- a one line change.

Fixes https://github.com/project-koku/koku-ui/issues/596
Fixes https://github.com/project-koku/koku-ui/issues/597

Network card:
Screen Shot 2019-03-12 at 3 06 20 PM

Database card:
Screen Shot 2019-03-12 at 3 08 38 PM


chargio:
Shouldn't the top services tab show Load Balancer-Network? it is showing EC2

chargio:
I am trying to visualize the price list, but I can't open it in an excel

PR #622, opened 2019-03-15
Changes to work under the cloud.redhat.com setup

chambridge:
FYI - I believe these are the set of changes for the upcoming app format changes to work on cloud.redhat.com.

FYI - @dlabrecq @boaz0 

Run from the insights proxy folder:
`SPANDX_CONFIG=/path/to/koku-ui/config/spandx.config.js bash scripts/run.sh`

then `yarn start:dev`

Point browser to: 
https://ci.foo.redhat.com:1337/apps/cost-management/

This should also work; but doesn't seem to:
https://ci.foo.redhat.com:1337/hcm/cost-management/

Worked of the changes seen here:
https://github.com/RedHatInsights/insights-frontend-starter-app/pull/62/files

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,16 +23,8 @@ User interface for Koku based on Patternfly [![Patternfly][pf-logo]][patternfly]
 From the `insights-proxy` project directory run the following command to interact with the deployed environment:
 
 ```
-bash scripts/run.sh
-```
-
-To run the koku-ui against a local koku server run the following command:
-
-```
 SPANDX_CONFIG=/path/to/koku-ui/config/spandx.config.js bash /path/to/insights-proxy/scripts/run.sh
 ```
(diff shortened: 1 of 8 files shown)


chambridge:
Think we'd just need another route here to support the `hcm/cost-management`:
https://github.com/project-koku/koku-ui/blob/master/src/index.tsx#L26

We are also going to have to update the api path:
https://github.com/project-koku/koku-ui/blob/master/src/api/api.ts#L21
I think its going to be:
`/api/cost-management/${version}/`

dlabrecq:
FYI, I have a workaround for the spandx.config.js file -- opened the Insights proxy issue below.

https://github.com/RedHatInsights/insights-proxy/issues/19

Issue #683, opened 2019-04-03
Ensure release name is not omitted from router path


dlabrecq:
**Describe the bug**
Ensure the release name (i.e., /beta) is not omitted from the router path per the Insights migration guide below.

https://docs.google.com/document/d/1Ku_MELkCjhY5ggwyQ7aawftiLMoKlqdpdsRPPSr8nyI/edit

**To Reproduce**
1. Go to 'Overview'
2. Add '/beta' to the browser URL (e.g., /beta/hybrid/cost-management)
3. See error

**Expected behavior**
The release should be included in the URL. Currently, "/beta/hybrid/cost-management" ends up as "/hybrid/cost-management".

**Screenshots**
![Screen Shot 2019-04-03 at 2 19 50 PM](https://user-images.githubusercontent.com/17481322/55502898-8fe89180-561b-11e9-9fa3-a73a4b741d05.png)

PR #684, opened 2019-04-03
Ensure release name is not omitted from router path

dlabrecq:
This should add the release name (e.g., /beta) into the router path.

Fixes https://github.com/project-koku/koku-ui/issues/683


dlabrecq:
Merging for QA

PR #714, opened 2019-04-06
Update spandx config

dlabrecq:
Update spandx config to include /app and /beta paths


dlabrecq:
Needed to continue local development

PR #776, opened 2019-04-19
fix(PriceList): update price list API and display

### LINKED ISSUE #765: Price List: API update
## User Story
As user, I want to have a friendly name to the metrics with the better units (GB-months/core-hours).

## Impacts
UI

## Role
User

## Assumptions
- [x] https://github.com/project-koku/koku/pull/817 is merged

## UI Details
https://redhat.invisionapp.com/share/HEO4ISHAT8Z#/screens/352926696

## Acceptance Criteria

- [x] Under metric use the display name and not the name with the first part of the unit
- [x] Rename `usage start` to `Applied usage range` and calculate its value by getting it from `usage_start` and `usage_end`
- [x] Rename `usage end` to `Applied usage date range`which is given in the second part of the unit

boaz0:
closes #765 

Update the Price List selectors to the new API.

- [x] ~~Fetch price list for only one cluster (optional)~~ - opened a follow up issue #780 
- [x] Show friendly name
- [x] Show tier number
- [x] Show applied usage range
- [x] Show applied usage range date

Questions:
- What should I do if both `usage start` and `usage end` are null or one of them? @nlcwong 

Screenshot

Screenshot_2019-04-19 Red Hat Cloud Software Services Cost Management

### DIFF
--- a/src/api/rates.ts
+++ b/src/api/rates.ts
@@ -15,7 +15,8 @@ export interface TieredRate {
 
 export type Rates = PagedResponse<Rate>;
 
-export function fetchRate() {
+export function fetchRate(uuid = null) {
+  const query = uuid ? `?provider_uuid=${uuid}` : '';
   const insights = (window as any).insights;
   if (
     insights &&
@@ -24,9 +25,9 @@ export function fetchRate() {
     insights.chrome.auth.getUser
   ) {
     return insights.chrome.auth.getUser().then(() => {
-      return axios.get<Rates>('rates/');
+      return axios.get<Rates>(`rates/${query}`);
     });
   } else {
-    return axios.get<Rates>('rates/');
+    return axios.get<Rates>(`rates/${query}`);
   }
 }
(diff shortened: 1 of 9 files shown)


dlabrecq:
Is this ready for review or are you still waiting for @nlcwong to answer your question?

Are you still implementing the "Fetch price list for only one cluster" feature?

boaz0:
@dlabrecq 

Yes, I am waiting on @nlcwong.
Regarding "fetch price list of one cluster" it depends how urgent it is - I believe it'll take me 3-4 days.

boaz0:
@dlabrecq 
I opened a new issue for "fetch price list of one cluster" as I think it will be too complicated in regards to caching and I think we should work on the design before implementing something that I'm not sure if it address all requirements that may come.
In addition, right now we have only 2-4 sources so it's not that bad, unless you say differently.

Thanks.

nlcwong:
@dccurtis when the usage start and usage end are null, does that mean the rate exists no matter what?  so like there is no tiering for this rate?

nlcwong:
@boaz0 assuming that null means that there is no usage range (meaning no tiers), then for now, put in "no range set"

boaz0:
Done
Screenshot from 2019-04-24 16-48-44

boaz0:
Thanks @dlabrecq

PR #777, opened 2019-04-21
fix(Source): update source settings table

### LINKED ISSUE #761: Sources: small adjustments
**Describe the bug**

- [x] Rename "last contracted" to "date added"
- [x] Rename "OCP" under type to "Red Hat OpenShift"
- [x] Move CluserID/ARN to underneath the name (in a tooltip explain that its the OpenShift Cluster ID or Amazon Resource Name)
- [x] Don’t have expansion row, just keep it as a data table list
- [x] Move the “added by” to its own column, and also maybe on tooltip (hover-over) show the email

**To Reproduce**
1. Click on "Add sources" in Overview

**Expected behavior**

**Screenshots**
https://redhat.invisionapp.com/share/XJR3A3996KR#/screens/356616244

boaz0:
closes #761 

screenshot


src/pages/sourceSettings/sourceSettings.tsx
```diff
@@ -121,40 +120,21 @@ class SourceSettings extends React.Component<Props, State> {
...
-          cells: [wrapper(src)],
-          // TODO: Uncomment when bulk delete is available
-          // selected: true,
-        },
-      ])
-      .reduce((acc, curr) => {
-        return [...acc, ...curr];
-      }, []);
+    const rows = sources.map((src, ix) => ({
+      cells: rowCell(t, src, () => {
```
dlabrecq: We should follow naming conventions and use camel case for classes. That is, `rowCell` should be `RowCell`.

It would be ideal if file names also match the class name.  For example, rename:

- `deleteDialog.tsx` as `messageDialogBase` (or rename the class as `DeleteDialog`)
- `detailsTable.styles.tsx` as `sourceTable.styles.tsx`
- `table.tsx` as  `sourceTable.tsx`
- `toolbar.tsx` as `filterToolbar.tsx` (or rename the class as just `Toolbar`)

src/pages/sourceSettings/sourceSettings.tsx
```diff
@@ -121,40 +120,21 @@ class SourceSettings extends React.Component<Props, State> {
     const columns = [
       t('source_details.column.name'),
       t('source_details.column.type'),
-      t('source_details.column.last_contacted'),
+      t('source_details.column.added_by'),
+      t('source_details.column.date_added'),
```
dlabrecq: Instead of providing `t` as a property, it would be preferred to wrap classes with `translate`. We do this in several classes like LoadingState, OcpReportSummary, and even MessageDialogBase, etc. 

We don't necessarily need to connect; for example, it can be as simple as:

`const OcpReportSummary = translate()(OcpReportSummaryBase);`

boaz0:
Thanks @dlabrecq

Issue #186, opened 2018-10-17
Replace FormGroup component with PF4


dlabrecq:
## User Story
Replace the custom FormGroup component with the PatternFly 4 Form/FormGroup components. This is currently being used in the providers modal, export modal, and login page.

## Impacts
UI

## Assumptions
The PatternFly 4 component can achieve similar visualizations. 

Note that the PF4 component does not appear to provide IDs.

## UI Details

The React based FormGroup component is now available for PatternFly 4.

http://patternfly-react.surge.sh/patternfly-4/components/form

## Acceptance Criteria

The modal achieves a similar layout with the new PF4 components.

dlabrecq:
Will modify the export modals. However, there is no need to modify the old login and providers modal. Those components are no longer in use.

PR #784, opened 2019-04-23
Replace custom components with PF4 components

dlabrecq:
Replaced custom FormGroup, TextInput, Link, and h2 tags with PF4 components. 

Note: Did not update the login page or old providers modal since those components are no longer in use. No change to the layouts.

Fixes https://github.com/project-koku/koku-ui/issues/186

Export modal:
Screen Shot 2019-04-23 at 2 24 24 PM

Filter text input:
Screen Shot 2019-04-23 at 4 09 04 PM


boaz0:
LGTM :+1:

PR #795, opened 2019-04-24
Chart keys

dlabrecq:
This ensures unique keys for each chart.

### DIFF
--- a/src/components/charts/costChart/costChart.tsx
+++ b/src/components/charts/costChart/costChart.tsx
@@ -263,7 +263,7 @@ class CostChart extends React.Component<CostChartProps, State> {
         <ChartArea
           data={datum.data}
           name={datum.name}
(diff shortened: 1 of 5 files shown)


boaz0:
@chargio this PR probably solves a warning message that React detected regarding key uniqueness. In general, it means that it will be hard for React to determine whether to re-render a specific part in the page or not. Using unique keys, React knows what changed and what not and by that gives good performances when refreshing or updating the page.

@dlabrecq LGTM :+1:  feel free to merge.

dlabrecq:
This is a technical issue with React. 

"Without a unique key, React couldn’t differentiate if the element was removed or just the content is changed."

You can read more about the technical issue here:
https://medium.com/@vraa/why-using-an-index-as-key-in-react-is-probably-a-bad-idea-7543de68b17c

PR #806, opened 2019-04-28
fix(Sources): update empty state texts to be more informative

### LINKED ISSUE #764: Sources: Empty screen text edits
**Describe the bug**
Text edits for the 'no sources configured' screen to provide more informative yet concise text.

**Expected behavior**
_In the grey space, please update the heading to:_

No sources are configured

_In the grey space, update the explanation to:_

To use Cost Management, you must connect the application to at least one source.

**Screenshots**
Screen 3 in https://redhat.invisionapp.com/share/XJR3A3996KR#/screens/356380301/comments/113855640 

**Additional context**
This wording is better for future translation, and provides more context to users.

Another option is to provide more info on this screen, like what Cloud Meter has done in screen #2 here: https://redhat.invisionapp.com/share/UFI078IZVAW#/screens/296433291. Thoughts?

boaz0:
closes #764 
//cc @dlabrecq @dayleparker

Screenshot_2019-04-28 Red Hat Cloud Software Services Cost Management

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -601,6 +601,12 @@
     "filter": {
       "placeholder": "Filter by source"
     },
+    "no_sources": {
+      "action": "Add source",
+      "desc": "To use Cost Management to manage your cloud costs, you must connect the application to at least one data source.",
+      "sub_desc": "Add an AWS account or an OpenShift Container Platform cluster to get started.",
+      "title": "No sources are configured"
+    },
     "remove_source": "Remove source",
     "title": "Cost management sources",
     "type": {
--- a/src/pages/sourceSettings/noSourcesState.styles.ts
+++ b/src/pages/sourceSettings/noSourcesState.styles.ts
@@ -1,0 +1,10 @@
+import { StyleSheet } from '@patternfly/react-styles';
+
+export const styles = StyleSheet.create({
+  container: {
+    display: 'flex',
+    justifyContent: 'center',
+    height: '100vh',
+    marginTop: '150px',
+  },
+});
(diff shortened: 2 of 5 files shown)


dayleparker:
Looks good @boaz0 👍

dayleparker:
Hey @boaz0 and @dlabrecq, 
It looks like this issue (#806) and #802 are for the same empty screen -- I apologize if my creating issues in GitHub has caused duplication. I'll double check the other ones I created last week...

Either screen is fine with me - @nlcwong, do you think one flow/wording is preferable over the other?
Cheers,
Dayle

boaz0:
I think the overview and the source settings empty screens should be different. Nevertheless, I don't mind using overview empty screen as that's what we have right now.
@dayleparker @nlcwong

dayleparker:
Thanks @boaz0 ! Apologies, I thought they were the same screen. If Natalie and Dan are OK with it, I say let's use them both in their respective spots in the UI, and review again later.

dlabrecq:
@dayleparker #806 and #802 are different screens. #802 is when shown when the user navigates to the overview and there are no providers. We don't show the tabs here, but an empty state which has a "view all sources" link, which navigates the user to the sources page.

They could be the same message, but only the sources page launches the providers modal.

boaz0:
Thanks @dlabrecq

nlcwong:
@dayleparker what @dlabrecq said is true.  The two screens are coming from two different places.  The settings section is really "platform" level, so the wording on the Cost Mgmt. overview page should redirect the user to the settings' cost sources page to then launch the wizard from there.  If we launched the wizard straight from the overview page, there would be no guidance for the user to know how to navigate to the sources page, they need to see that it in order to add any sources, they have to navigate to the settings section first (which makes it feel that it is "outside" of the cost application).  This will help setup the next phase of sources where we are all joined, outside of any one particular application.

PR #824, opened 2019-04-30
fix(Onboarding): update the first step instruction text

### LINKED ISSUE #787: Sources: Add a source > Select a source edits
**Describe the bug**
Text edits to the "Select a source" screen to make it clear and concise for users.

**Expected behavior**
Please update the instructional text to say:

Enter a descriptive name to identify the source, and select a source type.

**Screenshots**
https://redhat.invisionapp.com/share/XJR3A3996KR#/screens/356380290

**Additional context**
Before this text, this screen used to also have an initial sentence that said "To import data for cost management, connect to a data source." Is that something we want to keep, or do we want this UI screen to be usable for other apps?

boaz0:
closes #787 

//cc @dayleparker


boaz0:
Thanks @dlabrecq

Issue #846, opened 2019-05-02
AWS overview network and database card values don't add up


dchorvat1:
**Describe the bug**
Data for "top services" tab in network and database cards in AWS overview seem incorrect

**To Reproduce**
1. Go to AWS overview
2. See network/database services

**Expected behavior**
Values adding up to the totals

**Screenshots**
![Screenshot from 2019-05-02 11-43-50](https://user-images.githubusercontent.com/46324552/57067785-46e63480-6cd0-11e9-81d1-7f84460677fe.png)
![Screenshot from 2019-05-02 11-43-54](https://user-images.githubusercontent.com/46324552/57067787-49488e80-6cd0-11e9-979c-4a9b46d5a310.png)

**Desktop (please complete the following information):**
 - OS: RHEL
 - Browser Chrome
 - Version 73

**Additional context**
tested on CI beta

dlabrecq:
Using the wild card with `group_by[service]=*`, appears to take precedence over 

`filter[service]=AmazonRDS,AmazonDynamoDB,AmazonElastiCache,AmazonNeptune,AmazonRedshift,AmazonDocumentDB`

Replacing the group_by's wild card with the filter string seems to work, but only for the service tab. 

The other tabs appear to still need the wild card (e.g., `group_by[accounts]=*` and `group_by[regions]=*`) along with the filter for the costs to add up correctly.

dlabrecq:
It may be possible to back out our special-case code after the koku issue below has been addressed?

https://github.com/project-koku/koku/issues/848

PR #847, opened 2019-05-02
AWS overview network and database card values don't add up

dlabrecq:
AWS network & database card queries must use group_by instead of a filter for the service tab only.

Note that other tabs appear to still need the wild card (e.g., `group_by[accounts]=* `and `group_by[regions]=*`) along with the current filter for the costs to add up correctly.

Also need to ensure we don't apply logical AND to the group_by here.

Fixes https://github.com/project-koku/koku-ui/issues/846

Top services:
Screen Shot 2019-05-02 at 11 31 06 AM

Top regions:
Screen Shot 2019-05-02 at 11 32 52 AM

### DIFF
--- a/src/api/awsQuery.ts
+++ b/src/api/awsQuery.ts
@@ -39,7 +39,7 @@ const groupByAnd = 'and:';
 
 // Adds logical AND to group_by -- https://github.com/project-koku/koku-ui/issues/704
 export function getGroupByAnd(query: AwsQuery) {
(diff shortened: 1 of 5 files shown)


nbon12:
@dlabrecq I think the easiest way to handle this would be to throw an unsupported exception if you try to combine group_by=* with filter[service]=<name>, since what you're really trying to do is group_by=<name>, not group_by=*

Issue #649, opened 2019-04-01
Sources filter functionality


dlabrecq:
**Describe the bug**
There are some issues with filtering for the new sources page.

1. There is no results row when a filter is applied. As a user, I'm expecting to clear filters similarly to the details pages. Currently, the only way to clear a filter is to delete the previous filter string and press enter.

2. The filter does not accept partial strings. This could be an issue for the back-end API to solve. (In that case, please file an issue with koku.) However, I could not match "hccm", only exact matches (e.g., "hccm-vmware") are found.

**To Reproduce**
1. Go to 'Overview'
2. Click on 'Add Sources'
3. Enter a filter such as 'hccm'

**Expected behavior**
As a user, I should be able to clear individual filters. I should also be able to add partial filters.

**Screenshots**

Partial string:
![Screen Shot 2019-04-01 at 9 47 59 AM](https://user-images.githubusercontent.com/17481322/55332396-4ead8180-5463-11e9-8db1-d0e822f1a408.png)

Full String:
![Screen Shot 2019-04-01 at 9 48 14 AM](https://user-images.githubusercontent.com/17481322/55332426-5a00ad00-5463-11e9-8408-5e9e58b781dc.png)

**Additional context**
See https://docs.google.com/document/d/1_t1dV00l9j529vSWduzAkqtnGlLtlFE7e8zidylkGKo/edit?ts=5ca4eef3

boaz0:
@nlcwong @dlabrecq @dayleparker

Current state:
- [x] filter does accept partial strings: `name=boaz` will return any source that contains "boaz" in its name
- [x] filter does "AND" on type and name: `name=boaz&type=OCP` will return all OpenShift sources that contain "boaz" in their name
-[ ] filter does "OR" on name/type: `name=[boaz,david]` will return all sources that their names contain "boaz" or "david" (e.g. david-1, 1-boaz-david, david-boaz, boaz-23, davidistheman).

PR #855, opened 2019-05-05
feat(sources): add pagination to the table

### LINKED ISSUE #694: Sources: Add pagination
## User Story
As a user, I want to set the amount of items to display per page and to go to pages.

## Impacts
UI

## Role
Any user that has permissions to look at the sources

## Assumptions
None

## Acceptance Criteria

- [ ] Be able to change number of items to display per page
- [ ] Read results from other pages

boaz0:
closes #694

### DIFF
--- a/src/pages/sourceSettings/sourceSettings.tsx
+++ b/src/pages/sourceSettings/sourceSettings.tsx
@@ -20,6 +20,7 @@ import Header from './header';
 import { NoMatchFoundState } from './noMatchFoundState';
 import NoSourcesState from './noSourcesState';
 import RowCell from './rowCell';
+import SourcePagination from './sourcePagination';
 import { styles } from './sourceSettings.styles';
 import SourceTable from './sourceTable';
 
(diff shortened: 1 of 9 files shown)


src/pages/sourceSettings/sourceSettings.tsx
```diff
@@ -20,41 +20,72 @@ import Header from './header';
...
+
+  public onSearch(searchQuery) {
+    const newQuery = { ...this.props.query, ...searchQuery };
+    const res = Object.keys(newQuery)
+      .filter(k => newQuery[k])
+      .reduce((acc, curr) => {
+        const currQuery = `${curr}=${newQuery[curr]}`;
+        return acc === null ? currQuery : `${acc}&${currQuery}`;
+      }, null);
+    this.props.fetch(res);
```
dlabrecq: Is there any help text to suggest that the filter must be an exact match?

When filtering by type, I expect to see a filter results row and filter chips, similar to the details pages. 

I also expect this to be a partial, logical AND filter. For example, users should be able to enter "Red" and see all partial matches. However, I only see an unexpected error.

Screen Shot 2019-05-06 at 11 25 54 AM

FYI, I cannot get out of this state, unless I delete the filter text and hit enter.

boaz0: This addresses only #694, I can add filtering but the amount of files and code to review will be much bigger.

If that's not a problem on your side, let me know so I will continue working on this.

boaz0: In addition, the error you get is irrelevant to `exact match`. 
The backend will throw an error if the type filter is not one of ["OCP", "AWS'].
I suggest to replace it to selector when `Type`

dlabrecq: We can merge as is and address #649 separately. I was just confused by the filtering changes here.

boaz0:
Thanks @dlabrecq

Issue #856, opened 2019-05-06
Filter does not apply logical and to the results


dchorvat1:
**Describe the bug**
Logical and does not get applied to the filter results in details pages 

**To Reproduce**
1. Go to any details page
2. Filter by suitable values
3. See that the results are filtered by logical or instead of logical and

**Expected behavior**
Items filtered using logical and 

**Screenshots**
![Screenshot from 2019-05-06 10-17-48](https://user-images.githubusercontent.com/46324552/57214051-f7ae4580-6fe8-11e9-8096-966346bd1607.png)

**Desktop (please complete the following information):**
 - OS: RHEL
 - Browser Chrome
 - Version 73

PR #858, opened 2019-05-06
Apply logical AND for multiple group_by query params

dlabrecq:
Logical AND does not get applied to the filter results in details pages.

We don't want logical AND applied to the AWS overview network card, but expect to see the `and:` prefix when multiple group_by params are applied in the details pages. For example:

`group_by[and:project]=open&group_by[and:project]=sdn&order_by[cost]=desc`

Fixes https://github.com/project-koku/koku-ui/issues/856

Screen Shot 2019-05-06 at 4 50 39 PM


boaz0:
LGTM :+1:

Issue #851, opened 2019-05-03
kebab 'All projects' button shows all cluster projects rather than projects from selected cluster


lcouzens:
**Describe the bug**
kebab 'All projects' button shows all cluster projects rather than projects from selected cluster

**To Reproduce**
1. Go to 'Openshift Details'
2. Click on group by cluster
3. Click on kebab 'View all projects'
4. Modal shows all projects from all clusters

**Expected behavior**
We should only see selected projects from current cluster

**Screenshots**
![Screenshot from 2019-05-03 13-04-40](https://user-images.githubusercontent.com/18352403/57136883-107ae900-6da6-11e9-9c96-45db1621ea18.png)
![Screenshot from 2019-05-03 13-11-57](https://user-images.githubusercontent.com/18352403/57136893-183a8d80-6da6-11e9-926c-01ba1ca1a6e7.png)

**Desktop (please complete the following information):**
 - OS: Fedora
 - Browser Firefox
 - Version 66

**Additional context**
Testing against PROD with demo user

You can see from the screenshots that all the project data is the same for all 3 clusters

Looking at the API call for collecting the projects looks like we are collecting all instead of using multiple group bys cluster and project

`https://cloud.redhat.com/api/cost-management/v1/reports/openshift/costs/?filter[time_scope_units]=month&filter[time_scope_value]=-1&filter[resolution]=monthly&group_by[project]=*`

PR #859, opened 2019-05-06
'All projects' shows all cluster projects

dlabrecq:
The 'All projects' link and kebab menu open shows all cluster projects instead of the projects belonging to just that cluster.

Fixes https://github.com/project-koku/koku-ui/issues/851

Screen Shot 2019-05-06 at 5 28 35 PM


boaz0:
LGTM :+1:

Issue #849, opened 2019-05-03
Details page expanded row tabs change data on multple expanded rows


lcouzens:
**Describe the bug**
Details page expanded row tabs change data on multple expanded rows

**To Reproduce**
1. Go to 'cloud details'
2. Expand two rows
3. Click on regions tab on first row
4. Click on regions tab on second expanded row
5. see first expanded row data changes

**Expected behavior**
Expanded rows should be independent from each other

**Screenshots**
![Screenshot from 2019-05-03 10-27-36](https://user-images.githubusercontent.com/18352403/57130725-56c64d00-6d92-11e9-9f8d-94d1b20f7224.png)

**Desktop (please complete the following information):**
 - OS: Fedora
 - Browser Firefox
 - Version 66

**Additional context**
Testing against PROD with demo user

Issue #848, opened 2019-05-03
Line items on expanded row do not add to total on details page expanded row


lcouzens:
**Describe the bug**
Line items on expanded row do not add to total on details page expanded row

**To Reproduce**
1. Go to Details page
2. Expand any row
3. Add up line items for services or regions
4. Calculation does not add up to expanded row total 

**Expected behavior**
Items should add up to the total

**Screenshots**
![Screenshot from 2019-05-03 10-26-16](https://user-images.githubusercontent.com/18352403/57129976-fcc48800-6d8f-11e9-81ab-758f0f4291dd.png)

**Desktop (please complete the following information):**
 - OS: Fedora
 - Browser Firefox
 - Version 66

**Additional context**
Tested in PROD with demo user

PR #861, opened 2019-05-08
Manage tabs locally

dlabrecq:
The expanded row tabs have an affect on the tabs in other expanded rows. 

The problem occurs when changing tabs using Redux, while other rows are expanded. Basically, all rows are using the same Redux widget. This works well on the overview page because all cards are using unique widgets for their tabs. The fix here is to remove the Redux code in favor of managing tabs locally in the details page. 

In addition, this also ensures that we show only data associated with each row. For example, we were showing services for all projects instead of only projects for a given row. The fix was to add a filter for the current account, service, etc.

Fixes https://github.com/project-koku/koku-ui/issues/849
Fixes https://github.com/project-koku/koku-ui/issues/848

Before:
chrome-capture

After:
chrome-capture (1)


boaz0:
LGTM :+1:

Issue #863, opened 2019-05-08
Update charts for PF4 breaking changes


dlabrecq:
PatternFly is introducing some breaking changes to its charts. 

The first of many changes is removing the default 8 pixel padding. When this is in place, users cannot see both x and y axes. Our chart layouts do not have axis tics and labels on both axes; thus, we rely on the 8 pixel padding.

We will need to create our own Victory based theme in order to override the default padding and restore our chart layouts.

Overview charts:
![Screen Shot 2019-05-08 at 3 47 47 PM](https://user-images.githubusercontent.com/17481322/57403911-be243880-71a8-11e9-9b67-e2b04ef68191.png)

Bullet charts:
![Screen Shot 2019-05-08 at 3 48 09 PM](https://user-images.githubusercontent.com/17481322/57403903-b6649400-71a8-11e9-8f58-aeb7a162c0ba.png)

PR #864, opened 2019-05-08
Update charts for PF4 breaking changes

dlabrecq:
PatternFly is introducing some breaking changes to its charts. 

The first of many changes is removing the default 8 pixel padding. When this is in place, users cannot see both x and y axes. Our chart layouts do not have axis tics and labels on both axes; thus, we rely on the 8 pixel padding.

We will need to create our own Victory based theme in order to override the default padding and restore our chart layouts.

Fixes https://github.com/project-koku/koku-ui/issues/863

Overview charts:
Screen Shot 2019-05-08 at 3 50 29 PM

Bullet charts:
Screen Shot 2019-05-08 at 3 50 14 PM


src/components/charts/chartTheme/theme-common.ts
```diff
@@ -0,0 +1,269 @@
+import { ChartTheme } from '@patternfly/react-charts';
+
+const defaultTheme = ChartTheme.light.blue;
+
+// Theme
+export default props => {
+  const {
+    // COLOR_AXIS_FILL,
```
boaz0: Should we keep those?

dlabrecq: I was keeping for a reference, but will clean this up a bit.

PR #866, opened 2019-05-12
feat(priceList): Fetch price list per cluster and cache it for 30 minutes

### LINKED ISSUE #780: PriceList: Fetch price list per cluster
## User Story
As user, I want to fetch only the price list of a specific cluser so that the results will be render as soon as possible.

Right now, we are fetching all clusters' price lists which may take longer than fetching only one cluster price list.

## Impacts
UI

## Role
User

## Assumptions
- There is only one user who can update the price list (No race condition)

## UI Details
None

## Acceptance Criteria

- [x] When clicking on "View price list" of cluster X, fetch only cluster X price list using the API
- [x] Cache the price list of that cluster for at least 2 minutes (and update it when it is changed by that specific user).

boaz0:
closes #780 

- [x] fetch price list rates per cluster
- [x] cache the results for 2 minutes to reduce number of requests to the backend (this can be changed)

### DIFF
--- a/src/pages/ocpDetails/priceListModal.tsx
+++ b/src/pages/ocpDetails/priceListModal.tsx
@@ -31,8 +31,14 @@ interface Props extends InjectedTranslateProps {
 
 class PriceListModalBase extends React.Component<Props> {
   public componentDidUpdate() {
-    const { fetch, isOpen, providers, priceListStatus: status } = this.props;
-    if (isOpen && status === FetchStatus.none) {
+    const {
+      fetch,
+      isOpen,
+      providers,
+      priceListStatus: status,
+      name,
+    } = this.props;
+    if (isOpen && status !== FetchStatus.inProgress) {
       const priceListProvider = providers.data.find(p => p.name === name);
       fetch(priceListProvider ? priceListProvider.uuid : null);
     }
@@ -87,16 +93,21 @@ class PriceListModalBase extends React.Component<Props> {
 }
 
 const PriceListModal = connect(
-  createMapStateToProps(state => ({
-    priceList: priceListSelectors.ratesPerProvider(state),
-    priceListError: priceListSelectors.error(state),
-    priceListStatus: priceListSelectors.status(state),
-    providers: providersSelectors.selectProviders(
+  createMapStateToProps((state, props: { name: string | number }) => {
+    const providers = providersSelectors.selectProviders(
       state,
       ProviderType.ocp,
       'type=OCP'
-    ),
-  })),
+    );
+    const priceListProvider = providers.data.find(p => p.name === props.name);
+    const providerUuid = priceListProvider ? priceListProvider.uuid : null;
+    return {
+      priceList: priceListSelectors.ratesPerProvider(state, providerUuid),
+      priceListError: priceListSelectors.error(state, providerUuid),
+      priceListStatus: priceListSelectors.status(state, providerUuid),
+      providers,
+    };
+  }),
   {
     fetch: priceListActions.fetchPriceList,
   }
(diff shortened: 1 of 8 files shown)


src/pages/ocpDetails/priceListModal.tsx
```diff
@@ -30,11 +30,21 @@ interface Props extends InjectedTranslateProps {
...
-    if (isOpen && status === FetchStatus.none) {
+  public shouldComponentUpdate(nextProps) {
+    if (nextProps.isOpen !== this.props.isOpen) {
+      return true;
+    }
+    if (nextProps.priceListStatus !== this.props.priceListStatus) {
+      return true;
+    }
+    return false;
+  }
```
dlabrecq: A word of caution using shouldComponentUpdate. If anything is updated (e.g., if we later add inline editing, pagination, etc.), it will need to be checked here or componentDidUpdate/render won't be called as expected. I found this to be problematic elsewhere in the koku UI and since removed it.

src/store/priceList/actions.ts
```diff
@@ -14,12 +16,20 @@ export const {
...
+    const rates = cache[providerUuid];
+    const now = new Date();
+    if (rates && now.valueOf() - rates.time.valueOf() < TTL) {
+      dispatch(fetchPriceListSuccess(rates.data));
+      return;
+    }
     return fetchRate(providerUuid)
       .then(res => {
         dispatch(fetchPriceListSuccess(res.data));
+        cache[providerUuid] = { data: res.data, time: new Date() };
```
dlabrecq: Please take a look at how we fetch report in `awsReportsActions`, `ocpReportsActions`, etc. Although I didn't create the code initially, I believe that's a best practice approach for caching data.

In that code, we don't cache in awsReportsActions, we simply skip making a request for another report. Ideally, the reducer would cache the report, allowing the selector to return any stored report based on the `providerUuid` -- similar to how we provide unique IDs based on the widget ID / query.

Note that the `awsReportsActions` code below fetches a new report every 30 minutes.

```
export function fetchReport(
  reportType: AwsReportType,
  query: string
): ThunkAction<void, RootState, void, any> {
  return (dispatch, getState) => {
    if (!isReportExpired(getState(), reportType, query)) {
      return;
    }
...
```
    }

```
function isReportExpired(
  state: RootState,
  reportType: AwsReportType,
  query: string
) {
  const report = selectReport(state, reportType, query);
  const fetchStatus = selectReportFetchStatus(state, reportType, query);
  if (fetchStatus === FetchStatus.inProgress) {
    return false;
  }

  if (!report) {
    return true;
  }

  const now = Date.now();
  return now > report.timeRequested + expirationMS;
}
```

While the `awsReportsSelectors` code is able to obtain any cached report based on a unique ID / query.

```
export const selectReport = (
  state: RootState,
  reportType: AwsReportType,
  query: string
) => selectReportsState(state).byId.get(getReportId(reportType, query));
```

Finally, the `awsReortsReducer` code caches the report so the selector can retrieve by ID.

```
case getType(fetchAwsReportSuccess):
      return {
        ...state,
        fetchStatus: new Map(state.fetchStatus).set(
          action.meta.reportId,
          FetchStatus.complete
        ),
        byId: new Map(state.byId).set(action.meta.reportId, {
          ...action.payload,
          timeRequested: Date.now(),
        }),
        errors: new Map(state.errors).set(action.meta.reportId, null),
      };
```

src/pages/ocpDetails/priceListModal.tsx
boaz0: I can remove it. Don't know why I used it in the first place.

boaz0:
Thanks so much @dlabrecq 
I appreciate the time and help!

Issue #853, opened 2019-05-03
Incorrect empty state message when filtering on details pages


lcouzens:
**Describe the bug**
Incorrect empty state message when filtering on details pages

**To Reproduce**
1. Go to details page
2. Filter by 'test'
3. See message

**Expected behavior**
We should see a searching period, then a error message such as 'No clusters found with that filter'

**Screenshots**
![Screenshot from 2019-05-03 13-34-19](https://user-images.githubusercontent.com/18352403/57137681-7b2d2400-6da8-11e9-89a9-a8283a697753.png)

**Desktop (please complete the following information):**
 - OS: Fedora
 - Browser Firefox
 - Version 66

**Additional context**
Testing against PROD with demo user

PR #890, opened 2019-06-04
Empty state message when filtering

dlabrecq:
Adds an empty state message when filtering.

Fixes https://github.com/project-koku/koku-ui/issues/853

Screen Shot 2019-06-03 at 9 31 10 PM


boaz0:
Just a question.
In the source setting I am using also an empty state if there is no match. Should I use this instead?

dlabrecq:
I'll update the component (in another PR) to be more in line with yours. Then, you'll be able to use it as well.

Issue #893, opened 2019-06-05
Update empty filter state for use with sources


dlabrecq:
Update empty filter state for use with sources. For example we should use the same icons. We may also need a different description for the sources table.

PR #894, opened 2019-06-05
Updated empty filter state for use with sources

dlabrecq:
Updates the empty filter state component for use with the sources page. This ensures we use the same icons., but allows a different description for the sources table.

Fixes https://github.com/project-koku/koku-ui/issues/893

Screen Shot 2019-06-05 at 2 18 29 PM


boaz0:
Good job @dlabrecq

PR #908, opened 2019-06-26
fix(onboarding): update ocp and aws providers once on source creation

### LINKED ISSUE #907: Sources: Adding a new source causes API a DDoS attack on backend
**Describe the bug**
I noticed today while playing with onboarding that after adding a new source, the UI sends many requests to `api/v1/proviers?type=OCP` and ``api/v1/proviers?type=AWS`. It does that non-stop unless I'm refreshing my page.

**To Reproduce**
1. From the `overview` page expand `Settings`
2. Click on `Cost Management Sources`
3. In the `Settings` page click on `Add source`
4. Open Dev Tools (on Firefox it is done by clicking on F12) and go to the Network tab
5. Add a new source (follow the instruction. Note: It's easier to add a new OCP source than AWS)
6. After the source was added, notice that there are many requests sent to the backend and they are not stopping.

**Expected behavior**
No DDoS attack. I assume there should be only 1-2 call(s) to the backend.

**Screenshots**

boaz0:
fixes #907 
In general, I added another check that makes sure koku-ui won't sent more than 1 request per provider type when onboarding is completed successfully. 
By using prev props I can tell if a new source was added succesfully and in that moment to send a request.

### DIFF
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -107,16 +107,17 @@ export class App extends React.Component<AppProps, AppState> {
       onboardingStatus,
     } = this.props;
 
+    const wasOnboardingUpdated: boolean =
+      prevProps.onboardingStatus !== onboardingStatus &&
+      onboardingStatus === FetchStatus.complete;
     if (
-      (!awsProviders ||
-        (onboardingStatus === FetchStatus.complete && !onboardingErrors)) &&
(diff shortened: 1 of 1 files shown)


boaz0:
Thanks @dlabrecq

PR #914, opened 2019-07-10
fix(config): set the right routes to local koku from production

boaz0:
In order to run the local koku UI and local koku backend with the insights platform you need to set the route `api/cost-management` to the local koku backend and route `{/beta}*/{hybrid/api}/cost-management` to the insight proxy.

This PR updates `config/spandx.config.local.js` with those settings.


boaz0:
Thanks @dlabrecq !

PR #936, opened 2019-07-29
feat(onboarding): show a dialog to confirm  the user wishes to quit

### LINKED ISSUE #918: RFE: Sources UI: Add a 'are you sure you want to cancel?' failsafe
Can we provide a failsafe/extra check when the user hits Cancel? (ie. a popup window like "Are you sure you want to cancel adding this source?") 

It looked like a lot of info to collect while adding an OpenShift source, and could be a real pain for the user to retrieve all these values again if they hadn't noted the details somewhere else, and accidentally hit the Cancel button.

Not sure if we already have this in motion, but here's an idea I had after watching the Bluejeans recording (see the 14/06/19 BJN recording #2 at 1:09:00) for adding an OCP source.

boaz0:
closes #918 

image

### DIFF
--- a/src/store/onboarding/reducer.ts
+++ b/src/store/onboarding/reducer.ts
@@ -8,6 +8,8 @@ import {
   cancelOnboarding,
   checkSourceKindCheckList,
   closeModal,
+  displayConfirm,
+  hideConfirm,
   openModal,
   updateArn,
   updateClusterID,
@@ -27,6 +29,8 @@ export type Actions = ActionType<
   | typeof openModal
   | typeof closeModal
   | typeof cancelOnboarding
+  | typeof displayConfirm
+  | typeof hideConfirm
   | typeof checkSourceKindCheckList
   | typeof addSourceRequest
   | typeof addSourceSuccess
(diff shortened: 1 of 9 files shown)


src/store/onboarding/reducer.ts
```diff
@@ -147,6 +153,10 @@ export function reducer(state: State = defaultState, action: Actions): State {
       return { ...state, isOpen: true };
     case getType(closeModal):
       return { ...state, isOpen: false };
+    case getType(displayConfirm):
+      return { ...state, isOpen: false, isConfirmShown: true };
+    case getType(hideConfirm):
+      return { ...state, isOpen: true, isConfirmShown: false };
```
dlabrecq: Curious why this confirmation was added to the Redux store?

We've added a couple modals to Redux, so they could be launched from anywhere in the UI. However, the onboarding modal could easily display its own confirmation dialog.

boaz0: Showing two modals at the same time makes the confirmation dialog to be hidden behind the wizard modal. To fix this, I had to hide the wizard modal. If the confirmation modal `isOpen` state isn't stored in Redux then I will have to pass two onClose functions. One for closing the wizard modal and the other to show the confirmation modal.

To me it feels cleaner to put them in one place instead of passing two props that may sound the same.

dlabrecq: Ok, that sounds reasonable. Thanks for the explanation.

boaz0:
Thanks @dlabrecq

Issue #928, opened 2019-07-18
PF4 chart colors


dlabrecq:
PatternFly 4 charts are now themed with colors from pf-core. For consistency, we should consider removing the custom colors we created (for charts and legends) in favor of the themed colors provided by PatternFly.

boaz0:
//cc @aagarw10

dlabrecq:
Note that we will still need to customize the `colorScale` used by the chart component. That is, the colors representing the current and previous months. However, those colors should be pulled from react-tokens. 

We want to duplicate the same gray theme color pallet used by the PF4 chart. The green colors should come from the PF4 chart's green theme color pallet. And the `colorScale` should be applied in the same order.

PR #948, opened 2019-08-15
fixed issue #928 pf4 chart colors

aagarw10:
included commented out custom colors to show change to default imported pf4 core chart colors for bullet, cost, historical cost, historical trend, historical usage, pie, trend, and usage charts

### DIFF
--- a/src/components/charts/bulletChart/bulletChart.styles.ts
+++ b/src/components/charts/bulletChart/bulletChart.styles.ts
@@ -1,8 +1,5 @@
 import { StyleSheet } from '@patternfly/react-styles';
 import {
-  global_danger_color_100,
-  global_spacer_md,
-  global_spacer_sm,
   chart_color_black_100,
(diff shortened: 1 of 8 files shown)


dlabrecq:
The bullet chart range should use the gray scale.

Screen Shot 2019-08-15 at 4 50 55 PM

dlabrecq:
Never mind about the bullet chart using gray. We're no longer showing capacity, which is what had used the gray scale.

dlabrecq:
Looks like there is an issue that prevents Travis from building your PR. 

I created https://github.com/project-koku/koku-ui/pull/954 as a test. I included your changes, along with some modifications of my own. 

The problem appears to be with the comments included in the CSS style properties. The linter also requires inputs to be ordered alphabetically.

dlabrecq:
I've pushed a fix which should get the Travis build going

Issue #961, opened 2019-08-28
Add Azure details page and overview


dlabrecq:
We would like to add an Azure tab to the overview page, and an Azure details page. Until the restful APIs are available, this will just be a copy of AWS for now.

A separate Insights issue will be created to add a link to the details page.

PR #962, opened 2019-08-28
Added Azure details page and overview

dlabrecq:
Would like to get a jump on Azure and create an initial details page and overview. Until the restful APIs are available, this will just be a copy of AWS for now.

A separate Insights issue will be created to add a link to the details page.

Fixes https://github.com/project-koku/koku-ui/issues/961

Screen Shot 2019-08-27 at 10 32 28 PM


src/routes.tsx
```diff
@@ -11,6 +11,10 @@ const AwsDetails = asyncComponent(() =>
   import(/* webpackChunkName: "aws" */ './pages/awsDetails')
 );
 
+const AzureDetails = asyncComponent(() =>
+  import(/* webpackChunkName: "aws" */ './pages/azureDetails')
```
boaz0: I think this should change to

```js
   import(/* webpackChunkName: "azure" */ './pages/azureDetails')
```

PR #975, opened 2019-09-10
fix(costmodels): add scoped patternfly packages to the vendors bundle

### LINKED ISSUE #957: Cost model details page crashes after refresh
**Describe the bug**
After refreshing cost model details, the page crashes

**To Reproduce**
1. Navigate to cost model details
2. Refresh page
3. See error

**Expected behavior**
Page should display correctly

**Screenshots**
*Browser console output:*
Screenshot from 2019-08-23 12-32-27
*The page:*
Screenshot from 2019-08-23 12-32-36

**Desktop (please complete the following information):**
 - OS: rhel
 - Browser chrome
 - Version 75

**Additional context**
Tested and reproduced on both CI and QA

boaz0:
closes #957 

**Note**:
To verify this PR fixes the problem run `yarn start --env production` and go to /beta/hybrid/cost-management.

The `--env production` will create a minified version that is similar to what is used in the QA/CI environment (paas.psi.redhat.com).
I also set `hot: !isProduction` because in production the hot-module-reload is not used and can make the bundle not to be served.

//cc @dchorvat1 
//ping @karelhala can you review this too?

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -152,7 +152,7 @@ module.exports = env => {
       splitChunks: {
         cacheGroups: {
           commons: {
-            test: /[\\/]node_modules[\\/](react|react-dom|redux)[\\/]/,
+            test: /[\\/]node_modules[\\/](react|react-dom|redux|@patternfly*)[\\/]/,
             name: 'vendor',
             chunks: 'all',
           },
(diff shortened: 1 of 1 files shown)


webpack.config.js
```diff
@@ -152,7 +152,7 @@ module.exports = env => {
       splitChunks: {
         cacheGroups: {
           commons: {
-            test: /[\\/]node_modules[\\/](react|react-dom|redux)[\\/]/,
+            test: /[\\/]node_modules[\\/](react|react-dom|redux|@patternfly*)[\\/]/,
```
karelhala: Good catch! This means that if you are building commons package it will include `@patternfly` in it and load it correctly in minified version. Otherwise the treeshaking could remove this package from build.

Issue #965, opened 2019-09-03
Add tertiary navigation for AWS and Azure details


dlabrecq:
We want to use the PF4 navigation component as 3rd level navigation to the AWS and Azure details pages.

We should also remove "details" from the page title and just use "Amazon Web Services" and "Azure".

https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/navigation-system#horizontal-navigation

PR #977, opened 2019-09-10
Add tertiary navigation for AWS and Azure details

dlabrecq:
Per my discussion with @nlcwong , we want to use the PF4 navigation component as 3rd level navigation to the AWS and Azure details pages.

We're also removing "details" from the page title and just using "Amazon Web Services" and "Azure".

Fixes https://github.com/project-koku/koku-ui/issues/965

Screen Shot 2019-09-10 at 4 25 12 PM

### DIFF
--- a/src/components/details/tertiaryNav.tsx
+++ b/src/components/details/tertiaryNav.tsx
@@ -1,0 +1,109 @@
+import {
+  Nav,
+  NavItem,
+  NavList,
+  NavVariants,
+  Title,
+  TitleSize,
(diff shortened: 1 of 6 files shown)


boaz0:
Just need a clarification - when I am switching between AWS and Azure I get the same account names/services/etc. 
Is that OK or it's something you'll address in the future due to limitations on the back-end?


Thanks.

dlabrecq:
Yes, this is expected. Azure is just a copy of AWS right now. I will update the APIs when they become available next sprint

PR #970, opened 2019-09-09
fix(costmodels): change provider_uuids to providers in cost models api

boaz0:
closes #969

PR #979, opened 2019-09-11
fix(costmodels): show available sources in the source step

### LINKED ISSUE #978: in create cost models wizard - sources aren't shown even though they exist
**Describe the bug**
In selecting sources to a cost model, after data is fetched no sources are displayed even though there are sources.

**To Reproduce**
1. Click on 'Cost models details' in the sidebar
3. Enter name and description and select AWS as source type (you can select OCP but it's easier to reproduce with AWS)
4. Click next
5. Click next
6. See error

**Expected behavior**
See all the sources that are of the source type given in the first step.

**Screenshots**
image

boaz0:
this fixes the issue where no sources are displayed even though
they exist in the source step.

the issue was due to a change in the API of cost models:
the table was relaying on "provider_uuids" that was replaced
by "providers" and because of that the source list passed to
the table was empty.

closes #978 

related to #970 

//cc @chargio

### DIFF
--- a/src/pages/createCostModelWizard/api.ts
+++ b/src/pages/createCostModelWizard/api.ts
@@ -21,8 +21,8 @@ export const fetchSources = ({ type, page, perPage, query }) => {
     })
     .then(({ costmodels, sources }) => {
       const cmsHash = costmodels.reduce((acc, curr) => {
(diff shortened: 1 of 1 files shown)


boaz0:
Thanks @dlabrecq

PR #1002, opened 2019-09-18
fix(costmodels): update text

boaz0:
@jlalbertson @chargio these are the changes you asked.

- "Unassign" instead of "Delete"
image
- Cost model description header
image
- Use "per" instead of "for every"
image


boaz0:
Thanks @jlalbertson

PR #1005, opened 2019-09-20
fix(costmodel): display rates values correctly

### LINKED ISSUE #1004: The rate in dollar is not show correctly
**Describe the bug**
A rate is not displayed correctly:
 - if the rate is 10 then it's shown `$10` instead of `$10.00`
 - if the rate is 10.2 then it's shown `$10.2` instead of `$10.20`

**To Reproduce**
1. Go to 'Cost models details'
2. Click on any cost model
3. Add a new rate if there is none or edit an existing rate
4. Enter the values above
5. See error 

**Expected behavior**
10 => 10.00
10.2 => 10.20
10.32 => 10.32
10.432 => 10.432
and so on.

boaz0:
closes #1004 

image


boaz0:
Thanks @jlalbertson

PR #1009, opened 2019-09-22
feat(costmodel): add markup tab

boaz0:
Now that markup were added to the API, the UI can display, edit and delete a markup and also when creating a cost model, markup is included in the request.

**Note**
~~For some reason negative value for markup is not acceptable on server side.
I will open a ticket for this.~~
~~See ticket https://github.com/project-koku/koku/issues/1148~~

Fixes https://github.com/project-koku/koku-ui/issues/1017


dlabrecq:
Hold for tech exchange

"no changes should go in until after the EMEA conference. Once the EMEA conference is done, we'll want to get the cost model markup work in and available for the Americas Tech Exchange."

PR #1014, opened 2019-09-25
fix(costmodel): address warning messages in the console log

### LINKED ISSUE #1010: Error/warning log in console in cost models page
**Describe the bug**
In development mode there are error and warning messages regard missing key/aria-label prop in some element.

**Expected behavior**
Although these messages don't mean that the system will crash immediately they do alert about a problem that will cause a failure in the future.

**Screenshots**
Screenshot from 2019-09-23 17-15-25

Screenshot from 2019-09-23 17-15-55

Screenshot from 2019-09-23 17-16-10

boaz0:
closes #1010

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -319,6 +319,7 @@
       "pagination_section_aria_label": "bottom toolbar section - pagination",
       "results_text": "Results {{num}}",
       "select_filter_type": "Select filter type",
+      "table_aria_label": "assign sources to cost model table",
       "toolbar_results_section": "filter results section"
     },
     "steps": {
--- a/src/pages/costModelsDetails/addSourceStep.tsx
+++ b/src/pages/costModelsDetails/addSourceStep.tsx
@@ -126,6 +126,7 @@ class AddSourcesStep extends React.Component<AddSourcesStepProps> {
         </Toolbar>
         {sources.length > 0 && (
           <Table
+            aria-label={this.props.t('cost_models_details.add_source')}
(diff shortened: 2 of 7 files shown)


src/pages/costModelsDetails/costModelInfo/index.tsx
```diff
@@ -56,6 +56,7 @@ class CostModelInformation extends React.Component<Props, State> {
           <TabContent
             eventKey={0}
             id="refPriceList"
+            title="PriceList"
```
dlabrecq: Not sure how `TabContent` uses titles, but should be "Price list" for a title / aria attributes?

src/pages/costModelsDetails/costModelInfo/header.tsx
```diff
@@ -72,7 +72,7 @@ class Header extends React.Component<Props> {
         >
           <Tab
             eventKey={0}
-            title="Price list"
+            title="Pricelist"
```
dlabrecq: This should be two words (i.e., what you had there originally).

src/pages/costModelsDetails/costModelInfo/index.tsx
dlabrecq: Actually, I don't see `title` as a valid property for `TabContent`. Should be able to just remove it?

boaz0:
@dlabrecq thanks for the review - fixed all the comments.

also closes #1034

boaz0:
Thanks @dlabrecq

PR #1038, opened 2019-10-03
fix(costmodels): general information step fit and finish

### LINKED ISSUE #1037: Cost models fit & finish: create cost model - select source step
**Describe the bug**
Fit and finish

**Expected behavior**
- [x] Description field should be a text area
- [x] Padding around fields
- [x] Length of fields shouldn't fill the space. Set them to the length of the longest value in the source type option.

**Screenshots**
image

boaz0:
closes #1037 

- [x] Description field should be a text area
- [x] Padding around fields
- [x] Length of fields shouldn't fill the space. Set them to the length of the longest value in the source type option.

image

### DIFF
--- a/src/pages/createCostModelWizard/generalInformation.tsx
+++ b/src/pages/createCostModelWizard/generalInformation.tsx
@@ -1,7 +1,11 @@
 import {
+  Form,
   FormGroup,
   FormSelect,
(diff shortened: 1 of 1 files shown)


boaz0:
Thanks @dlabrecq mind reviewing again?
image

PR #1040, opened 2019-10-03
fix(costmodels): add price list sub-step fit and finish

### LINKED ISSUE #1039: Cost models fit & finish: create cost model - price list sub step - add rate
**Describe the bug**

Fit and finish to the sub step in create cost model wizard.

**Expected behavior**

- [x] Missing spaces between fields
- [x] Use place holder instead of setting a default value: `0.00`
- [x] Add a dollar sign to the rate field
- [x] Set width of fields to at least 340px

**Screenshots**

image

boaz0:
closes #1039 

- [x] Missing spaces between fields
- [x] Use place holder instead of setting a default value: `0.00`
- [x] Add a dollar sign to the rate field
- [x] Set width of fields to at least 340px

image

### DIFF
--- a/src/pages/createCostModelWizard/addPriceList.tsx
+++ b/src/pages/createCostModelWizard/addPriceList.tsx
@@ -1,140 +1,171 @@
 import {
   Button,
+  Form,
   FormGroup,
   FormSelect,
   FormSelectOption,
+  InputGroup,
+  InputGroupText,
+  Stack,
+  StackItem,
+  Text,
+  TextContent,
   TextInput,
+  TextVariants,
   Title,
   TitleSize,
 } from '@patternfly/react-core';
+import { DollarSignIcon } from '@patternfly/react-icons';
+import { css } from '@patternfly/react-styles';
 import React from 'react';
 import { InjectedTranslateProps, translate } from 'react-i18next';
 import { CostModelContext } from './context';
 import { units } from './priceListTier';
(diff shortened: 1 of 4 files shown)


src/pages/createCostModelWizard/addPriceList.tsx
```diff
@@ -24,53 +28,18 @@ const AddPriceList: React.SFC<InjectedTranslateProps> = ({ t }) => {
...
-                  label={t('cost_models_wizard.price_list.memory_metric')}
-                />
-                <FormSelectOption
-                  value="storage"
-                  label={t('cost_models_wizard.price_list.storage_metric')}
-                />
-              </FormSelect>
-            </FormGroup>
-            {priceListCurrent.metric !== '' && (
+            <Form style={{ width: '350px' }}>
```
dlabrecq: Can we please create a CSS selector for this?

dlabrecq:
The title layout looks a little off. We're using PatternFly's stack component in the RBAC wizard, like so.

Screen Shot 2019-10-03 at 9 59 21 AM

This is the result:

Screen Shot 2019-10-03 at 9 58 33 AM

FYI, we're also using PatternFly's TextListItem component for the confirmation step. This nice thing about that is you don't need to deal with grids. The layout accommodates the longest label.

Screen Shot 2019-10-03 at 9 59 46 AM

PR #1065, opened 2019-10-19
fix(costmodels): details page sources tab fit and finish

### LINKED ISSUE #1059: Cost models fit and finish - cost model details page sources tab
**Describe the bug**
 Fit and finish to the sources tab in cost model details page.

**Expected behavior**
- [x] Assign new sources should be a modal and not a wizard
image

- [x] Delete dialog should be small dialog
image

- [x] Empty state: similar to the provider empty state (add -> assign)
image

boaz0:
closes #1059 

- [x] Assign new sources should be a modal and not a wizard
image

- [x] Delete dialog should be small dialog
image
**UPDATE**
image

- [x] Empty state: similar to the provider empty state (add -> assign)
image

### DIFF
--- a/src/components/filter/filterComposition.tsx
+++ b/src/components/filter/filterComposition.tsx
@@ -17,6 +17,7 @@ interface TypeOption {
 }
 
 interface Props extends InjectedTranslateProps {
+  isSingleOption?: boolean;
   options: TypeOption[];
   id: string;
   query: Query;
@@ -70,6 +71,7 @@ const FilterCompositionBase: React.SFC<Props> = ({
   updateFilter,
   switchType,
   onSearch,
+  isSingleOption = false,
   t,
 }) => {
   const filterController =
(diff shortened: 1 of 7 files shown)


boaz0:
//ping @dlabrecq @nlcwong please review this.

nlcwong:
For the delete dialog, we should just keep the button title generic (since in the wording it's pretty generic) that we say we are un-assigning a source rather than un-assigning a cluster.  @dayleparker, would you have thoughts on that?

boaz0:
I am adding @chargio to this conversation too.

dayleparker:
> For the delete dialog, we should just keep the button title generic (since in the wording it's pretty generic) that we say we are un-assigning a source rather than un-assigning a cluster. @dayleparker, would you have thoughts on that?

I agree @nlcwong -- the button should be just "Unassign source".

I think the wording in that dialog box could be a little more smooth/succinct if we removed the word "using" from the first sentence, so:

"This will remove the assignment of [SOURCE] from the [COSTMODELNAME] cost model."

One last comment - for the empty state message, do you think we need "like an Amazon Web Services account"? Or could this be explanatory enough with just "Assign a source to this cost model." ?

nlcwong:
@dayleparker I'm a bit flip floppy on this issue.  I think less is more, but in this case, the user who is setting up the cost model might not know what a source is because they might not be the one who added the source in the first place.  In fact i can see where certain sources will be delegated to different users (say a cluster belongs to engineering vs. IT ops and they all handle their pricing differently).  So the empty state might need to be more specific in that sense.

boaz0:
//ping @dlabrecq  - read for your review :raised_hands:

Issue #1066, opened 2019-10-21
Switch path from hybrid to cost-management


ryelo:
## User Story
Cost management is becoming its own bundle on cloud.redhat.com and moving out of the `/hybrid` path. The path should now be `/cost-management`

## Impacts
UI

## Acceptance Criteria
all references to `hybrid` should change to `cost-management`

## Refs
https://github.com/RedHatInsights/landing-page-frontend/pull/96
https://github.com/RedHatInsights/cloud-services-config/pull/82
https://github.com/RedHatInsights/insights-chrome/pull/412
https://github.com/RedHatInsights/entitlements-api-go/pull/34

dlabrecq:
I've updated our spandx config, which allows me to login. However, the provider APIs fail with 403. Thus, I cannot see the overview page, but an error message. And when I click on "OpenShift details"; for example, I lose the left-side nav. 

Overview error:
![Screen Shot 2019-10-21 at 1 50 21 PM](https://user-images.githubusercontent.com/17481322/67230155-c1d01280-f40a-11e9-9a93-af6520c674a0.png)

Navigation error:
![Screen Shot 2019-10-21 at 1 50 33 PM](https://user-images.githubusercontent.com/17481322/67230177-ca284d80-f40a-11e9-86cc-0564bc560c12.png)

PR #1067, opened 2019-10-22
Switch path from hybrid to cost-management

dlabrecq:
All references to `hybrid` should change to `cost-management`

https://github.com/project-koku/koku-ui/issues/1066

Doesn't address issues with left-side Insights nav


boaz0:
@ryelo - I can't navigate to settings. It redirects me to overview page.

dlabrecq:
The setting navigation has been fixed by Insights -- I've confirmed with CI beta

PR #1069, opened 2019-10-24
fix(navigation): use history push only for user actions

karelhala:
This will enable custom app navigation trough `insights.chrome.appNavClick({id: 'cost-management-sources', secondaryNav: true})` and not triggering redirect.

### DIFF
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -79,9 +79,11 @@ export class App extends React.Component<AppProps, AppState> {
       insights.chrome.navigation(buildNavigation());
     }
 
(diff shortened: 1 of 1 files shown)


boaz0:
@karelhala I was testing it but I am still redirected to the overview page.

karelhala:
This is only to activate the correct item in navigation. Since sources are managed by cost-management app we don't want to do full redirect. If you are in your app and there is some button `Take me to sources` you have to udpate history in order to react-router to take effect and call this to enable it in navigation. I am also working on counter PR in chrome to investigate why it's not properly highliting.

boaz0:
I see now. Thanks.

PR #1113, opened 2019-11-19
fix(costmodels): on submit click on the 'Next' button in the wizard

### LINKED ISSUE #1105: On clicking enter in text input the create cost model wizard reloads the page
**Describe the bug**
On creating a cost model through the wizard when choosing a text input and pressing Enter the wizard disappears and it looks like the cost models page is refreshing.

**To Reproduce**
1. Go to 'cost model details'
3. Focus on the Name field
4. Click Enter
5. See error

**Expected behavior**
When being on a text input and clicking Enter, click on the Next button to move to the next step if it's enabled.

boaz0:
closes #1105

### DIFF
--- a/src/components/forms/form.tsx
+++ b/src/components/forms/form.tsx
@@ -1,0 +1,24 @@
+import {
+  Form as FormPF4,
+  FormProps as FormPF4Props,
+} from '@patternfly/react-core';
+import React from 'react';
+import { Omit } from 'react-redux';
+
+type FormProps = Omit<FormPF4Props, 'onSubmit'>;
+
+export const Form: React.ComponentType<FormProps> = ({
+  children,
+  ...props
+}) => {
+  return (
+    <FormPF4
+      onSubmit={(event: React.FormEvent<HTMLFormElement>) =>
+        event.preventDefault()
+      }
+      {...props}
+    >
+      {children}
+    </FormPF4>
+  );
+};
--- a/src/pages/createCostModelWizard/addPriceList.tsx
+++ b/src/pages/createCostModelWizard/addPriceList.tsx
@@ -1,6 +1,5 @@
 import {
   Button,
-  Form,
   FormGroup,
   FormSelect,
   FormSelectOption,
@@ -17,6 +16,7 @@ import {
 } from '@patternfly/react-core';
 import { DollarSignIcon } from '@patternfly/react-icons';
 import { css } from '@patternfly/react-styles';
+import { Form } from 'components/forms/form';
 import React from 'react';
 import { InjectedTranslateProps, translate } from 'react-i18next';
 import { CostModelContext } from './context';
--- a/src/pages/createCostModelWizard/generalInformation.tsx
+++ b/src/pages/createCostModelWizard/generalInformation.tsx
@@ -1,5 +1,4 @@
 import {
-  Form,
   FormGroup,
   FormSelect,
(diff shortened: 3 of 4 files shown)


src/components/forms/form.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Omit } from 'react-redux';
+
+type FormProps = Omit<FormPF4Props, 'onSubmit'>;
+
+const ignoreSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+  event.preventDefault();
+  const submitButton = document.querySelector(
+    'button.pf-m-primary[type="submit"]'
+  ) as HTMLButtonElement;
+  submitButton.click();
```
dlabrecq: This makes the assumption that every form will have a button that matches the class 'button.pf-m-primary[type="submit"]'. Instead of making an assumption that button exists, it would be preferable to have the button's onClick call `preventDefault`.

For example, in ocpDetails/detailsTag.tsx, I do this in the button's onClick function.

```
public handleOpen = event => {
    this.setState({ isOpen: true });
    event.preventDefault();
    return false;
  };
```

And in azureDataSource/index.tsx, azureCreds/index.tsx, etc. I see that you have been doing this:

```
<Form
      onSubmit={e => {
        e.preventDefault();
        return false;
      }}
    >
```

boaz0: I am not sure I understood what you were saying.
If I'm not wrong when you're pressing Enter in the input box, it triggers the "submit" event.
Although I agree with you that I'm doing here a bad assumption, I don't think calling preventDefault from the onClick button will fix it. Did you try it?

boaz0: In addition, the requirement was to click on the Next button on pressing enter on the input box so I'm not sure how that works.

boaz0: Intresting....

boaz0: Unfortunately, in the wizard I can not set the "onClick" props of the buttons.
Nevertheless, even if it was possible, the problem was pressing enter when you're on the input box. But going to every text input and adding `onClick={e => e.preventDefault()}` doesn't sound so good.

I do understand that I'm doing an assumption of having the submit button. So what I can do is move this to `createCostModelWizard` instead putting it in `components`

dlabrecq: Pressing enter in the input box shouldn’t navigate the user to the next step. (Perhaps set focus on the button?) The user shouldn’t navigate elsewhere unless they click on the next button. Is that what’s happening?

boaz0: there is a conflict then. @nlcwong - in the fit & finish I remembered you mentioning on pressing enter in the input box was to move to the next button

image

dlabrecq: Looking at the PatternFly wizard example and the text input does not navigate the user to the next step. Although, that could be a PatternFly issue?

nlcwong: What i wrote on the box is the issue.  It's not supposed to behave that way.  When a user clicks the "Enter" key, it should not auto-advance the step UNLESS the focus is on the primary button.  If the focus is on the primary button, wouldn't any key then make the button depress and do that action?

boaz0: OK thanks.

boaz0:
ping @dlabrecq can you review this now?

PR #1114, opened 2019-11-21
fix(costmodels): Add timeout to markup validation in the wizard

### LINKED ISSUE #1106: Set a delay to the markup validation
**Describe the bug**
In the create cost model wizard, at the markup step if "-1" is typed then you will see immediately an error message saying that the value is invalid and then it will disappear. This makes the UI feel odd.

In order to fix this, adding a small delay (280 ms) on the validation may fix this.

**To Reproduce**
1. Go to 'Cost model details'
3. Set AWS as type and enter a name
4. Click Next
5. Type "-1" in the text box
6. See the error

**Expected behavior**
When typing -1, there should be no error message if the value is entered fast.

boaz0:
closes #1106

### DIFF
--- a/src/pages/createCostModelWizard/markup.tsx
+++ b/src/pages/createCostModelWizard/markup.tsx
@@ -15,51 +15,78 @@ import React from 'react';
 import { InjectedTranslateProps, translate } from 'react-i18next';
 import { CostModelContext } from './context';
 
-const Markup: React.SFC<InjectedTranslateProps> = ({ t }) => (
-  <CostModelContext.Consumer>
-    {({ onMarkupChange, markup }) => {
-      const isValidMarkup = !isNaN(Number(markup));
-      return (
-        <Stack gutter="md">
-          <StackItem>
-            <Title size="xl">{t('cost_models_wizard.markup.title')}</Title>
-          </StackItem>
-          <StackItem>
-            <TextContent>
-              <Text component={TextVariants.h6}>
-                {t('cost_models_wizard.markup.sub_title')}
-              </Text>
-            </TextContent>
-          </StackItem>
-          <StackItem>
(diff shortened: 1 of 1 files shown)


src/pages/createCostModelWizard/markup.tsx
```diff
@@ -15,51 +15,79 @@ import React from 'react';
...
+                        id="markup"
+                        name="markup"
+                        value={markup}
+                        onChange={(value: string) => {
+                          if (this.timer) {
+                            clearTimeout(this.timer);
+                          }
+                          this.timer = setTimeout(() => {
+                            this.setState({ isValid: !isNaN(Number(value)) });
+                          }, 280);
```
dlabrecq: Using timeouts is typically frowned upon because it's not best practice. Perhaps it would be better to perform the validation via onBlur, when the user leaves the input? 

With the onChange event, validation is performed with each key press, which isn't desirable.

boaz0: If we validate the value onChange the error message will still be shown as it does right now.
Using `onBlur` is also not perfect because if you type `minus` and focus on a different thing and then update to `-1` the error message will be shown unless you're focusing on a different element again.

dlabrecq: If you type minus, then leave the input, it would be expected to invalid. Regardless, my point is that this should be done without a timeout

boaz0: I see what you mean. OK - I will look at it.

boaz0: OK - I updated the code, hope you'll like it.

boaz0:
No problem - thanks for your time and patience

PR #1125, opened 2019-11-24
fix(costmodels): display warning when going to override assignment

### LINKED ISSUE #1119: Display warning when assigning a new cost model to a source that has a cost model
## User Story
As User, I want to see a warning on a source that has a cost model assigned to so that I know that the action I am doing is going to overwrite current settings.

## Acceptance Criteria

- [ ] If the source(s) selected were assigned to a different cost model beforehand, we should show a warning message icon and tooltip to let user know that one or more had been assigned to another cost model before and this will override that.
- [ ] Display a warning icon and a tooltip on both the assign source(s) and review steps. Only when a source is checked show a warning icon.

boaz0:
closes #1119 

s1

image

//cc @dayleparker

### DIFF
--- a/src/pages/createCostModelWizard/review.tsx
+++ b/src/pages/createCostModelWizard/review.tsx
@@ -22,6 +22,7 @@ import React from 'react';
 import { InjectedTranslateProps, translate } from 'react-i18next';
 import { CostModelContext } from './context';
 import { getLabels, PriceListTier } from './priceListTier';
+import { WarningIcon } from './warningIcon';
 
 const ReviewSuccessBase: React.SFC<InjectedTranslateProps> = ({ t }) => (
   <CostModelContext.Consumer>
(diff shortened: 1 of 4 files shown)


dayleparker:
Hi @boaz0,
This looks good! A few small things for the text:
**Screen 1:**
-  I think this screen is missing the instruction at the top - maybe something like "Select the source(s) to assign this cost model to."
- Popover text: Wording makes sense but the spelling for 'assignment' needs a fix :) Or another possibility for wording, with or without the cost model name in brackets -- "This selection will override the current cost model assignment (MyCost)." Either works I think - @dlabrecq any thoughts?

**Screen 2:**
Looks great but I suggest removing the quotation marks around "Create" and "Back". (Can it be bold text in Patternfly?)

Cheers,
Dayle

src/pages/createCostModelWizard/review.tsx
```diff
@@ -111,7 +112,14 @@ const ReviewDetailsBase: React.SFC<InjectedTranslateProps> = ({ t }) => (
...
                 <TextListItem component={TextListItemVariants.dt}>
-                  {t('cost_models_wizard.steps.sources')}
+                  {t('cost_models_wizard.steps.sources')}{' '}
+                  {sources.find(
+                    src => src.selected && src.costmodel !== undefined
+                  ) && (
+                    <WarningIcon
+                      text={t('cost_models_wizard.warning_override_sources')}
+                    />
+                  )}
```
dlabrecq: Wondering if the icon should appear on the right-most side? That is, instead of next to the label, located to the right of the value. Perhaps something to consider in a following PR?

Screen Shot 2019-11-29 at 6 48 45 PM

boaz0: Yes, @nlcwong and I were taking about this and probably there is going to be a follow up.
Thanks.

PR #1126, opened 2019-11-24
fix(costmodels): show confirmation dialog before closing wizard

### LINKED ISSUE #1120: Show confirmation dialog before quiting the create cost model wizard
## User Story
As User, I want to see a confirmation dialog before closing the wizard so that if I clicked "Cancel" by accident I won't lose my progress.

## Acceptance Criteria
- [ ] Show confirmation dialog when the user clicks "Cancel".
- [ ] The confirmation dialog should have two options: Yes and Cancel.
- [ ] Yes is for stopping the creation process
- [ ] Cancel is for getting back to the last point the user were in the wizard

boaz0:
closes #1120 

image

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -259,6 +259,11 @@
   },
   "cost_models_wizard": {
     "cancel_button": "Cancel",
+    "confirm": {
+      "close": "Yes, I want to cancel",
+      "message": "Are you sure you want to cancel the creation process? All settings will be discarded.",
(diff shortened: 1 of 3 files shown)


dayleparker:
Hi @boaz0, 
This text makes sense as it is (would just need a comma after "yes"), but I suggest shortening it a little to:

> Are you sure you want to cancel the creation process? All settings will be discarded.

What do you think of this wording?

PR #1215, opened 2020-01-08
refactor(costmodel): get price list metrics from API and use it in UI

### LINKED ISSUE #1213: Get price list rates' metrics and measurements from server
## User Story
As a developer, I want to get metrics and measurements for rates so that I won't need to update UI any time a new metric or measurement is added on the backend.

## Impacts
UI
## Acceptance Criteria

- [ ] Get metrics and measurements from `api/v1/metrics`

boaz0:
Refactored the whole cost models price list components to make it easier to use the same metrics data structure.

Read the price list metrics from API using Redux.

closes #1213


boaz0:
Thanks Dan!

PR #1264, opened 2020-01-20
Added webpack stats target

dlabrecq:
Added webpack stats target to view package structure. 

The generated stats file can be viewed below
http://chrisbateman.github.io/webpack-visualizer/


boaz0:
That's a great idea! :+1:

Issue #1266, opened 2020-01-21
[create cost model wizard] Allow user to go back to price list table


boaz0:
## User Story

As a user, I want to be able to go to price list table (when there is at least one rate) so that I won't have to create a new rate.

## Impacts
UI

## Acceptance Criteria

- [x] Show the "Cancel" button when adding additional rate.
- [x] Change the "Save rate" button text to "Add rate"
- [x] Make "Add rate" and "Cancel" next to each other

lcouzens:
Verified

PR #1269, opened 2020-01-23
feat(costmodel): add the cancel button to add rate form in wizard

boaz0:
related #1266 

- [x] Show the "Cancel" button when adding additional rate.
- [x] Change the "Save rate" button text to "Add rate"
- [x] Make "Add rate" and "Cancel" next to each other


nlcwong:
Would it be possible to see a screenshot of the change?

boaz0:
@nlcwong 
No rates:
image

Creating first rate:
image

Adding a second rate
image

image

dlabrecq:
Sounds like @nlcwong requested changes?

boaz0:
@dlabrecq - that's right.

boaz0:
I will update the ticket but till a new change is pushed, DO NOT MERGE

package.json
```diff
@@ -102,16 +102,20 @@
...
     "@babel/plugin-syntax-decorators": "^7.2.0",
     "@babel/plugin-syntax-dynamic-import": "^7.2.0",
     "@babel/plugin-syntax-jsx": "^7.2.0",
     "@babel/plugin-syntax-typescript": "^7.2.0",
+    "@testing-library/jest-dom": "^5.0.2",
+    "@testing-library/react": "^9.4.0",
     "@types/enzyme": "^3.1.15",
     "@types/enzyme-adapter-react-16": "^1.0.3",
     "@types/jest": "^23.1.1",
+    "@xstate/test": "^0.3.0",
```
dlabrecq: We will want to start locking down the dependency versions. Instead of `"@xstate/test": "^0.3.0"`, simply use "0.3.0"

nlcwong:
@boaz0 I'm a bit confused, I thought that when the user lands on the Price list step, the very first thing they would see is the rates table with an empty state in the middle:
EmptyTable

and then if they click on the "Add rate" button, they would get this screen:
AddRate

boaz0:
image

image

image

nlcwong:
@boaz0 I just noticed something in the screenshots, that I couldn't have noticed on my phone, the wizard buttons, they need to be disabled too while in "add rate" mode.  It is weird to have the back button and cancel button of the wizard to be enabled while having the cancel button enabled for the add rate mode.  Does that make sense?

boaz0:
I can not disable the cancel/back buttons but I can hide them.

image

dlabrecq:
@nlcwong I don't believe we should disable or hide the cancel and back buttons. It feels very odd that I cannot go back/cancel unless we finish adding a rate.

nlcwong:
PF comment was to hide the Next button as well.  In a later issue, I will request we use a drawer component to fit all this into (will have to mock up a design for that).

boaz0:
@nlcwong Done
image

Issue #1290, opened 2020-01-29
Make data toolbar more generic


dlabrecq:
Want to make the details data toolbar more generic so it can (possibly) be used in pages other than details. 

- Syncing groupBy with selected category options should be optional
- The export icon should be optional
- Filtering by tags should be optional
- Options should not be localized inside the data toolbar
- Restoring chip state based on filter_by query params should be optional
- The category options should be optional

dchorvat1:
verified, c9e8b13abe99720621407871a6268422044354d9

PR #1291, opened 2020-01-29
Make data toolbar more generic

dlabrecq:
Want to make the details data toolbar more generic so it can (possibly?) be used in pages other than details. 

- Syncing groupBy with selected category options should be optional
- The export icon should be optional
- Filtering by tags should be optional
- Options should not be localized inside the data toolbar
- Restoring chip state based on filter_by query params should be optional
- The category options should be optional

https://github.com/project-koku/koku-ui/issues/1290


dlabrecq:
FYI, Boaz.

Issue #1300, opened 2020-01-30
Cannot regenerate yarn.lock


dlabrecq:
With node 8, I'm seeing the following error while generating the yarn.lock file.

<img width="807" alt="Screen Shot 2020-01-30 at 9 14 28 PM" src="https://user-images.githubusercontent.com/17481322/73507046-a1b6d480-43a5-11ea-9029-fd41e74f2dfd.png">

Unfortunately, we cannot currently build with node >= 10. If I remove the `yarn.lock` and regenerate via `yarn install`, I too many errors.

dlabrecq:
jsdom@16.0.1 appears to be a dependency of react-merlin, which is currently being used for the onboarding modal.

<img width="897" alt="Screen Shot 2020-01-30 at 9 29 30 PM" src="https://user-images.githubusercontent.com/17481322/73507652-9fee1080-43a7-11ea-8f09-622ba5823c71.png">

dlabrecq:
I can lock the versions in package.json in order to install / build with node 10. However, I would also like to move the onboarding modal to the `archive/pages` directory?

Considering onboarding has moved to sources, there is no longer a need to build this. Thus, we can remove the dependency on `react-merlin`.

PR #1301, opened 2020-01-31
Fixed yarn.lock generation

dlabrecq:
Fixed yarn.lock generation -- tested with node 8 & 10. 

- Locked versions in package.json -- ensures pkgs remain compatible with node 8/10
- Removed `react-merlin` dependency, which pulls in `jsdom@16.0.0`
- Moved unused onboarding modal to archive dir, which depends on `react-merlin`
- Moved unused login session code to archive dir -- we use Insights to login
- Fixed tests by adding testURL to jest config
- Added a couple types, which allows building/running with node 10

Note: All changes above were required to install, build, and run tests properly with node 8 & 10.

To test, delete the `yarn.lock` file and run `yarn install`

https://github.com/project-koku/koku-ui/issues/1300


boaz0:
I guess this is going to conflict with https://github.com/project-koku/koku-ui/pull/1269 (xstate, xstate/test and react testing library)

dlabrecq:
@boaz, this shouldn't conflict unless the dependencies you're adding depend on node >= 10?

Instead of using `"@testing-library/jest-dom": "^5.0.2"`, simply specify that as "5.0.2"

boaz0:
LGTM :+1:

PR #1311, opened 2020-02-06
Update to handle special case where tag key collides with keywords.

chambridge:
* Utility function to see if "tag:" prefixed entry exist on tag group_by
* Update report items logic to use utility function
* Update SPANDX config for running local koku with latest insights-proxy
* Update README with overview link

This handles the following special case:
https://github.com/project-koku/koku/pull/1690

Tag_Special_case

### DIFF
--- a/src/utils/getComputedOcpReportItems.ts
+++ b/src/utils/getComputedOcpReportItems.ts
@@ -6,6 +6,7 @@ import {
   OcpReportValue,
 } from 'api/ocpReports';
 import { Omit } from 'react-redux';
+import { getItemLabel } from './getItemLabel';
 import { sort, SortDirection } from './sort';
 
 export interface ComputedOcpReportItem {
@@ -98,12 +99,13 @@ export function getUnsortedComputedOcpReportItems({
             : '';
         const id = `${value[idKey]}${idSuffix}`;
         let label;
-        if (labelKey === 'cluster' && cluster_alias) {
+        const itemLabelKey = getItemLabel({ report, labelKey, value });
+        if (itemLabelKey === 'cluster' && cluster_alias) {
           label = cluster_alias;
-        } else if (value[labelKey] instanceof Object) {
-          label = (value[labelKey] as OcpDatum).value;
+        } else if (value[itemLabelKey] instanceof Object) {
+          label = (value[itemLabelKey] as OcpDatum).value;
         } else {
-          label = value[labelKey];
+          label = value[itemLabelKey];
         }
         const limit = value.limit ? value.limit.value : 0;
         const request = value.request ? value.request.value : 0;
--- a/src/utils/getItemLabel.ts
+++ b/src/utils/getItemLabel.ts
@@ -1,0 +1,21 @@
+export interface GetItemLabelParams {
+  report: any;
+  labelKey: any;
+  value: any;
+}
+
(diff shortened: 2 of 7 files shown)


src/utils/getItemLabel.ts
```diff
@@ -0,0 +1,21 @@
...
+  value: any;
+}
+
+export function getItemLabel({ report, labelKey, value }: GetItemLabelParams) {
+  let itemLabelKey = String(labelKey);
+  if (report.meta && report.meta.group_by) {
+    const group_by = report.meta.group_by;
+    for (const key of Object.keys(group_by)) {
+      if (key.indexOf('tag:')) {
+        const tagPrefixKey = 'tag:' + labelKey;
```
dlabrecq: FYI, we have this tag prefix defined in Query.

https://github.com/project-koku/koku-ui/blob/master/src/api/query.ts#L4

dlabrecq:
@chambridge does this also cover the issue below? 
https://github.com/project-koku/koku/issues/1658

dlabrecq:
Do we also want to update getComputedOcpCloudReportItems.ts as well?

src/utils/getComputedOcpReportItems.ts
```diff
@@ -98,12 +99,13 @@ export function getUnsortedComputedOcpReportItems({
...
+        const itemLabelKey = getItemLabel({ report, labelKey, value });
+        if (itemLabelKey === 'cluster' && cluster_alias) {
           label = cluster_alias;
-        } else if (value[labelKey] instanceof Object) {
-          label = (value[labelKey] as OcpDatum).value;
+        } else if (value[itemLabelKey] instanceof Object) {
+          label = (value[itemLabelKey] as OcpDatum).value;
         } else {
-          label = value[labelKey];
+          label = value[itemLabelKey];
```
dlabrecq: I'm wondering if this is old, unused code because the debugger never seems to step into the `else if` statements?

It appears that the itemLabelKey could return something with "tag:" prefix. However, will the report "values" ever contain "tag:project", for example?

Screen Shot 2020-02-06 at 9 17 38 AM

src/utils/getItemLabel.ts
chambridge: Done. Thanks for the pointer 🙇

chambridge:
@dlabrecq Yes, the koku PR referenced above (project-koku/koku#1690) addresses issue: project-koku/koku#1658

src/utils/getComputedOcpReportItems.ts
chambridge: As mentioned here: https://github.com/project-koku/koku/pull/1690#issuecomment-582891216
This is special case logic to colliding keywords. The `tag:` prefix wouldn't be placed on projects, but it will now for a tag like `cost` -> `tag:cost`, so it doesn't collide with the total `cost` value.

If you reference the API output shown in the following link you can see the difference: https://github.com/project-koku/koku/pull/1690#issue-371042654

Hope that clarifies.

dlabrecq: Ok, I think I got it. You're adding the tag prefix if and only if there is a collision.

I see that collision (and prefix) with Azure after group_by[tag:cost] now.

chambridge: yep.

Issue #1322, opened 2020-02-11
No sources link should point the user to platform sources


chambridge:
**Describe the bug**

Now sources integration is complete our links should be pointing users to that UI (cost management sources should no longer be used).

https://github.com/project-koku/koku-ui/blob/cdf31e9f4f3c2b7f61d51cf1f12fe63b09514914/src/components/state/noProvidersState/noProvidersState.tsx#L32

<img width="1434" alt="image" src="https://user-images.githubusercontent.com/29379759/74252881-4f868500-4cbc-11ea-94a1-b15d8cea0ddb.png">

**To Reproduce**
1. Login to an account without sources created
2. Click on Cost Management
3. See link to setup sources and click

<img width="1426" alt="image" src="https://user-images.githubusercontent.com/29379759/74253130-aee49500-4cbc-11ea-8117-3f027a70a818.png">

**Expected behavior**
We should link users to platform sources (https://cloud.redhat.com/beta/settings/sources)
<img width="1428" alt="image" src="https://user-images.githubusercontent.com/29379759/74253198-c9b70980-4cbc-11ea-8ed6-c914c6f93b6e.png">

lcouzens:
My guess is this didn't land in CI yet for w/e reason, or its still not fixed.

lcouzens:
@dlabrecq could it be from the failed travis run? https://travis-ci.org/project-koku/koku-ui/builds/649700134?utm_source=github_status&utm_medium=notification

dlabrecq:
Yea, that's a likely problem. Will look into that, thanks.

lcouzens:
Verified

PR #1323, opened 2020-02-12
Fixed sources link and archived unused code

dlabrecq:
Fixed sources link for the "no providers" state. Also archived unused sources code.

https://github.com/project-koku/koku-ui/issues/1322


dlabrecq:
Snapshots appear to be ok

Issue #1306, opened 2020-02-05
Create cost model wizard (assign sources): PF data toolbar


boaz0:
Replace existing PatternFly toolbar, on the details pages, with the experimental PatternFly data toolbar. The PF data toolbar follows the latest PatternFly design pattern, which includes new icons, chip layout, etc.

dchorvat1:
I found some issues with `assign sources` step in create cost model wizard:
  - It seems to me that the toolbar is still the PatternFly Toolbar and not a PatternFly Data Toolbar
  - For Azure, this step misses some localized strings
  - This step could not display sources we can assign, even though there were providers present

Screenshot:
![Screenshot from 2020-03-03 11-13-06.png](https://images.zenhubusercontent.com/5c3320c5ea2ab97902b1b15b/89e5783f-79c0-415d-abc9-15c8c8de26ef)

cc @boaz0

boaz0:
@dchorvat1 - please be advised, you don't see the sources because the `/sources/` endpoint doesn't contain the "cost_models" attribute.

There is a PR for that https://github.com/project-koku/koku/pull/1832.

dchorvat1:
@boaz0 yep, that looks like it's been resolved. Besides that, could you take a look at the two other points I addressed?

boaz0:
@dchorvat1 - in regards to your other points:
- I checked it and it's using the Data Toolbar (you can see by the chips when you add query).
- I fixed the translation error in #1363

PR #1347, opened 2020-02-26
feat(costmodels): Migrate to PF4 DataToolbar in Assign Sources

boaz0:
address #1306

### DIFF
--- a/src/pages/createCostModelWizard/assignSourcesToolbar.tsx
+++ b/src/pages/createCostModelWizard/assignSourcesToolbar.tsx
@@ -1,0 +1,99 @@
+import {
+  InputGroup,
+  InputGroupText,
+  Pagination,
+  PaginationProps,
+  TextInput,
+} from '@patternfly/react-core';
+import {
+  DataToolbar,
+  DataToolbarContent,
+  DataToolbarFilter,
+  DataToolbarItem,
+} from '@patternfly/react-core/dist/esm/experimental';
+import { SearchIcon } from '@patternfly/react-icons';
+import React from 'react';
+import { InjectedTranslateProps, translate } from 'react-i18next';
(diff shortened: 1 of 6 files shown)


src/pages/createCostModelWizard/assignSourcesToolbar.tsx
```diff
@@ -0,0 +1,97 @@
...
+            />
+          </DataToolbarFilter>
+        </DataToolbarItem>
+        <DataToolbarItem variant="pagination">
+          <Pagination {...paginationProps} />
+        </DataToolbarItem>
+      </DataToolbarContent>
+    </DataToolbar>
+  );
+};
```
dlabrecq: Instead of having to provide `t` as a prop, you could do use `translate` instead. For example:

const AssignSourcesToolbar = translate()(AssignSourcesToolbarBase);
export { AssignSourcesToolbar };

src/pages/createCostModelWizard/table.tsx
```diff
@@ -57,38 +59,44 @@ const SourcesTable: React.SFC<InjectedTranslateProps> = ({ t }) => {
...
+                  isCompact: true,
+                  itemCount: sources.length,
+                  perPage,
+                  page,
+                  onSetPage: (_evt, newPage) => {
+                    fetchSources(type, query, newPage, perPage);
+                  },
+                  onPerPageSelect: (_evt, newPerPage) =>
+                    fetchSources(type, query, 1, newPerPage),
+                }}
```
dlabrecq: Just a suggestion... Instead of providing component props, you could provide the pagination component itself. This helped make my toolbar a little cleaner, didn't need to provide `t`, etc. Then, the table can interact with the pagination component's callbacks directly.

Issue #1346, opened 2020-02-25
Overview: Refactor Azure


dlabrecq:
Refactor the "Azure" overview page to use a dropdown in conjunction with tab navigation. This also removes progress bars for most cards (i.e., except 1st), and updates titles.

Mock: https://marvelapp.com/4b95ibf/screen/66066595

![Screen Shot 2020-02-25 at 5 02 10 PM](https://user-images.githubusercontent.com/17481322/75291633-97df9000-57f0-11ea-96bc-c393b9ea799b.png)

dchorvat1:
verified, 54ec1fef4d374c1453fdf1a996d591660260ff7c

Issue #1345, opened 2020-02-25
Overview: Refactor AWS


dlabrecq:
Refactor the "AWS" overview page to use a dropdown in conjunction with tab navigation. This also removes progress bars for most cards (i.e., except 1st), and updates titles.

Mock: https://marvelapp.com/4b95ibf/screen/66066593

![Screen Shot 2020-02-25 at 5 01 18 PM](https://user-images.githubusercontent.com/17481322/75291561-75e60d80-57f0-11ea-90d3-77d9f0d3fd60.png)

dchorvat1:
verified, 54ec1fef4d374c1453fdf1a996d591660260ff7c

Issue #1344, opened 2020-02-25
Overview: Refactor OpenShift on cloud infrastructure


dlabrecq:
Refactor the "OpenShift on cloud infrastructure" overview page to use a dropdown in conjunction with tab navigation. This also removes progress bars for most cards (i.e., except 1st), and updates titles.

Mock: https://marvelapp.com/4b95ibf/screen/65806666

![Screen Shot 2020-02-25 at 4 59 55 PM](https://user-images.githubusercontent.com/17481322/75291523-623aa700-57f0-11ea-8746-709373c8ba07.png)

dchorvat1:
verified, 54ec1fef4d374c1453fdf1a996d591660260ff7c

Issue #1343, opened 2020-02-25
Overview: Refactor OpenShift tab


dlabrecq:
Refactor the "OpenShift" overview page to use a dropdown in conjunction with tab navigation. This also removes progress bars for most cards (i.e., except 1st), and updates titles.

Mock: https://marvelapp.com/4b95ibf/screen/65806667

![Screen Shot 2020-02-25 at 4 59 16 PM](https://user-images.githubusercontent.com/17481322/75291428-2d2e5480-57f0-11ea-9ca8-512079af98ef.png)

dchorvat1:
verified, 54ec1fef4d374c1453fdf1a996d591660260ff7c

PR #1349, opened 2020-02-27
Refactor overview tabs

dlabrecq:
Refactored the "OpenShift on cloud infrastructure" overview page to use a dropdown in conjunction with tab navigation. This also removes progress bars for non-cost summary cards and updates titles.

Mocks: https://marvelapp.com/4b95ibf/screen/65806667

Note that the following new pages are awaiting new APIs -- see  https://github.com/project-koku/koku/issues/1762. These pages show data from existing APIs for now.

- Ocp supplementary
- Aws filtered by Ocp
- Azure filtered by Ocp
- Ocp usage

chrome-capture

See
https://github.com/project-koku/koku-ui/issues/1343
https://github.com/project-koku/koku-ui/issues/1344
https://github.com/project-koku/koku-ui/issues/1345
https://github.com/project-koku/koku-ui/issues/1346


src/locales/en.json
```diff
@@ -579,22 +625,22 @@
...
   },
   "ocp_cloud_dashboard": {
-    "compute_title": "Cloud infrastructure compute services usage",
+    "compute_title": "Compute services usage",
     "compute_trend_title": "Daily usage comparison ({{units}})",
     "cost_label": "Cost",
-    "cost_title": "OpenShift cloud infrastructure cost",
-    "cost_trend_title": "Cloud infrastructure cumulative cost comparison ({{units}})",
+    "cost_title": "All infrastructure filtered by OpenShift cost",
+    "cost_trend_title": "All OpenShift related infrastructure cumulative cost comparison ({{units}})",
```
ddonahue007: Just curious. Why would you have the ned for the 'All' and 'filtered by' here? It just seems strange to have search and filter key words in the title. Nothing to hold up the PR.

dlabrecq: Good question. The short answer is; this matches Natalie's mocks. We may need to wordsmith this a bit...

I believe "all" is used to convey that this includes OpenShift, Aws, and Azure.

Mock:
Screen Shot 2020-02-28 at 11 44 04 AM

ddonahue007: Good point. If the data set is made up of a culmination of providers then It makes sense. I guess I was not looking at that way.

Issue #1332, opened 2020-02-20
cost model details (price list tab): PF data toolbar


boaz0:
Replace existing PatternFly toolbar, on the details pages, with the experimental PatternFly data toolbar. The PF data toolbar follows the latest PatternFly design pattern, which includes new icons, chip layout, etc.

dchorvat1:
verified

PR #1355, opened 2020-03-02
feat(costmodels): migrate to pf4 datatoolbar in price list tab

boaz0:
related #1332 
image

image

### DIFF
--- a/src/pages/costModelsDetails/components/priceListTable.tsx
+++ b/src/pages/costModelsDetails/components/priceListTable.tsx
@@ -1,6 +1,4 @@
 import {
-  Button,
-  Chip,
   DataList,
   DropdownItem,
   EmptyState,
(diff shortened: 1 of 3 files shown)


src/pages/costModelsDetails/components/priceListTable.tsx
```diff
@@ -211,96 +268,82 @@ class PriceListTable extends React.Component<Props, State> {
...
-                  >
-                    {this.state.filter}
-                  </Chip>
-                )}
-              </ToolbarItem>
-            </ToolbarGroup>
-          </ToolbarSection>
-        </Toolbar>
+        <PriceListToolbar
+          t={t}
```
dlabrecq: We shouldn't have to provide `t` if the base components each use `translate`.

Issue #1358, opened 2020-03-03
No cost data available even though there are reports and providers.


boaz0:
**Describe the bug**
Even though, there are providers with data when going to Azure or to OCP there is a message saying: `No cost data available yet`.

**To Reproduce**
1. Go to 'OpenShift cloud infrastructure details'
2. See error

**Expected behavior**
Display table with details

**Screenshots**
![image](https://user-images.githubusercontent.com/2453279/75764461-8f360f00-5d46-11ea-888b-0e3745b1a2fa.png)

**Additional context**
Looking at the Network tab, all requests are successful especially to `/reports/`.
I don't see requests to `/providers/` though.

dlabrecq:
The API recently changed from `providers` to `sources`. This may not be available in CI, yet?

See https://github.com/project-koku/koku-ui/pull/1350/files

dlabrecq:
This appears to be working for me, running local CI beta, using the cost-demo user.

The following APIs are returning data as expected.

https://ci.foo.redhat.com:1337/api/cost-management/v1/sources/?type=AWS
https://ci.foo.redhat.com:1337/api/cost-management/v1/sources/?type=AZURE
https://ci.foo.redhat.com:1337/api/cost-management/v1/sources/?type=OCP

![Screen Shot 2020-03-03 at 9 59 49 AM](https://user-images.githubusercontent.com/17481322/75788153-d071f300-5d35-11ea-849b-f29dd9d5f630.png)

![Screen Shot 2020-03-03 at 9 59 58 AM](https://user-images.githubusercontent.com/17481322/75788162-d4057a00-5d35-11ea-8c65-1dcfcc6da8ea.png)

![Screen Shot 2020-03-03 at 10 00 09 AM](https://user-images.githubusercontent.com/17481322/75788172-d8ca2e00-5d35-11ea-85bf-42843df93dd5.png)

dlabrecq:
Confirmed this is working with CI beta and the cost-demo user.

boaz0:
//cc @dchorvat1 can you look at this since you mentioned this to me.

PR #1350, opened 2020-02-28
Switch from providers to sources API endpoint as providers will be removed.

chambridge:
* Removing providers endpoint soon to avoid confusing with sources
* Sources endpoint was updated to provide the same support the providers endpoint had.

Image shows local run with calls to the Sources API
image

PR #1359, opened 2020-03-03
fix(details): display no cost data available yet when no reports

boaz0:
related #1358 

@dlabrecq can you look at this. I was looking into it earlier and thought maybe I can help.
Let me know if it looks good and if it's not close to fix it, then feel free to close and open your own PR.

Thanks.


dlabrecq:
This looks like a dup of 1358.

The API recently changed from providers to sources. This may not be available in CI, yet?

See https://github.com/project-koku/koku-ui/pull/1350/files

dlabrecq:
I believe this should be closed. I've confirmed that CI beta is working with the `sources` API.

Issue #1338, opened 2020-02-24
UI User can add new cost models and edit them in the interface


chargio:
**Describe the bug**
A user is capable of following the Ui to create or modify cost models, even if he has no rights to do so.

**To Reproduce**
1. Go to 'Cost models'
2. Click on 'Create new model'
3. Follow the workflow and select a source

**Expected behavior**
The platform shows a "this feature is read only". I was expecting all the buttons to be disabled if the user is read only (no create new cost model, you can see the model but you can't modify, delete or add rates)

**Screenshots**
![image](https://user-images.githubusercontent.com/7500590/75162612-17505f00-571e-11ea-883e-8f6725f91e56.png)

**Desktop (please complete the following information):**
 - OS: Mac OS X
 - Browser [Firefor, chrome, safari]
 - Version [latest]

**Additional context**
It looks like the user can edit things he can't

boaz0:
OK - so I couldn't reproduce this anymore on production but I figured out where I can get retrieve this information.

Using chrome insights:

```
return insights.chrome.auth.getUser().then((user) => {
  const isOrgAdmin = user.identity.user.is_org_admin;
  // ... do something with this information ...
})
```

Probably storing this in Redux will be useful in the future.

nlcwong:
Affected actions (in primary buttons & in kebabs options): 
- Create cost model
- Delete cost model
- Add rate
- Delete rate
- Edit markup
- Assign source
- Unassign source

Expected behavior: 
In addition to enable/disable button/links, include a tooltip with following text:
"You do not have permission to create or delete a cost model."
"You do not have permission to add, edit, or remove a rate."
"You do not have permission to assign or unassign a source."

@adberglund @chambridge  Can the UI call the RBAC APIs directly to check to see if the user has permissions for cost model CRUD?  If so, what would that API be for us to check to see if a button/action/link should be enabled/disabled due to permissions?

adberglund:
@boaz0 @nlcwong did you see this https://github.com/RedHatInsights/insights-chrome/blob/master/src/js/entry.js#L125-L127 is that useful here? I don't think knowing if they're an org admin is good enough here.

boaz0:
@adberglund that's what I thought but that returns `true` or `false` and @nlcwong gave me the impression it's more granular than that.

boaz0:
[...]

PR #1367, opened 2020-03-09
feat(costmodel): render read-only view

boaz0:
related to #1338 

image

image

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "@redhat-cloud-services/frontend-components": "^1.0.8",
     "@redhat-cloud-services/frontend-components-notifications": "^1.0.1",
     "@redhat-cloud-services/frontend-components-utilities": "^1.0.0",
+    "@redhat-cloud-services/rbac-client": "^1.0.53",
     "@types/date-fns": "2.6.0",
     "@types/i18next": "8.4.4",
     "@types/i18next-xhr-backend": "1.4.1",
--- a/src/api/rbac.test.ts
+++ b/src/api/rbac.test.ts
@@ -1,0 +1,7 @@
+import { getRBAC } from './rbac';
+
+test('getRBAC', () => {
+  getRBAC().then(resp => {
+    expect(resp).toEqual({ isOrgAdmin: false, permissions: null });
+  });
+});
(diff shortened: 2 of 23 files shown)


boaz0:
DO NOT MERGE YET... missing unit tests

boaz0:
//cc @nlcwong - are you sure you want to add tooltip to each of the disabled items?

nlcwong:
Yes.  So the plan is that we are going to go ahead with this approach and use it to test it out on our users to see if they get it.  Worse case, they don't get it and we'll need to figure out an alternative, but until that testing (we are thinking "summit" timeframe) let us go with this approach.

boaz0:
@nlcwong @dlabrecq - there is a problem to get tooltips into dropdown items because it wraps the component inside "disabled" item. So it's not perfect but it's better than nothing.

Screenshot from 2020-03-12 11-26-17

Screenshot from 2020-03-12 11-26-26

Screenshot from 2020-03-12 11-27-24

dlabrecq:
Do we really want disabled tooltips on individual dropdown options? IMO, no.

Currently, we don't show tooltips for enabled dropdown options, nor does PatternFly demo a pattern in their examples.

nlcwong:
Boaz,
Does all the items in the drop down need to be disabled for that tooltip to
appear or just one?  I'll forward this to Margot and Kendra to figure this
out from the RH Cloud perspective.

Dan,
I get your concern and Matt Carrano has the same concern w/r/t discovery
and whether or not this should be a PF guidance.  The game plan is to go
ahead with the tooltip and test with users.  The RBAC team is having the
same issues and actually this whole disabled thing was born out of that
discussion.  I know Kendra is trying to start a thread going to find more
use cases from the other products on how they are handling things and will
work with Matt Carrano to get an official guidance on this.  Until then,
we're going to go ahead with the disabled button tooltip approach since
that feels the simplest (since launching a dialog that is a notification
would be more involved) and we want to test something out.

- Natalie

NATALIE WONG

SENIOR UX INTERACTION DESIGNER

Red Hat <https://www.redhat.com/>

300 A Street, Boston, MA 02210

nwong@redhat.com    T: +1-857-770-0627    M: +1-508-963-6144
<https://red.ht/sig>

dlabrecq:
I'm ok with the disabled button, but strongly against adding this to individual dropdown options.

boaz0:
Found a workaround

Screenshot from 2020-03-17 00-02-37

Screenshot from 2020-03-17 00-02-50

nlcwong:
MIND BLOWN!

PR #1382, opened 2020-03-17
Updates to handle new cost terminology API changes.

chambridge:
* Changes to consume API changes
* Does not handle perspectives or just showing supplemental or infrastructure


archive/components/reports/ocpCloudReportSummary/ocpCloudReportSummaryDetails.tsx
```diff
@@ -41,7 +41,6 @@ const OcpCloudReportSummaryDetailsBase: React.SFC<OcpCloudReportSummaryDetailsPr
 }) => {
   let cost: string | React.ReactNode = <EmptyValueState />;
   let infrastructureCost: string | React.ReactNode = <EmptyValueState />;
-  let markupCost: string | React.ReactNode = <EmptyValueState />;
```
boaz0: Files that are under the archive directory are obsolete and meant to be reference.

src/components/reports/reportSummary/reportSummaryDetails.tsx
```diff
@@ -119,8 +121,8 @@ const ReportSummaryDetailsBase: React.SFC<ReportSummaryDetailsProps> = ({
           content={t(
             `${perspective}_dashboard.total_cost_tooltip`,
             perspective === DashboardPerspective.ocp
-              ? { infrastructureCost, derivedCost }
-              : { infrastructureCost, markupCost }
+              ? { infrastructureCost, supplementaryCost }
+              : { infrastructureCost, supplementaryCost }
```
dlabrecq: We were showing derived and markup costs on different pages (e.g., Ocp on cloud showed markup). Just want to ensure markup is no longer available and they're both supplementary now?

See https://github.com/project-koku/koku-ui/blob/ffbcf10cf807d5d001d4b92ec372154638fac127/src/locales/en.json#L646

For the Ocp overview, the tooltip that shows:

Screen Shot 2020-03-17 at 2 35 06 PM

For the Ocp on cloud overview (now "All infrastructure filtered by Ocp cost"), the tooltip that shows:

Screen Shot 2020-03-17 at 2 35 29 PM

dlabrecq: If markup is removed, then we don't need that `perspective === DashboardPerspective.ocp` code

dlabrecq:
@nlcwong 

For the Ocp overview page, we have a tooltip to show the infrastructure and derived cost.
For the Ocp on cloud overview page (aka All filtered by Ocp cost), we have a tooltip to show the infrastructure and markup cost.

With this change, both tooltips show infrastructure and supplementary cost. Is that what we want?

archive/components/reports/ocpCloudReportSummary/ocpCloudReportSummaryDetails.tsx
chambridge: Sure. I can put the files back but the git repo commits keep the history, so you could just delete this folder then and refer to your SCM.

src/components/reports/reportSummary/reportSummaryDetails.tsx
chambridge: So based on our revision to the cost terminology markup is not removed but can be part of both infrastructure and supplementary: https://docs.google.com/document/d/1hnwS0uwPuAKw75QyE2QEl8IZcczaQredMvDY10UVhLk/edit?usp=sharing

I didn't make adjustments to the perspectives, because I think the code flow needs to change more to pass the perspecive to the `runReport` method so that you can properly choose whether you want to use the following APIs:
- /reports/openshift/costs/
- /reports/openshift/infrastructures/all/costs/         (openshift on all infrastructures)
- /reports/openshift/infrastructures/aws/costs/      (openshift on all aws)
- /reports/openshift/infrastructures/azure/costs/      (openshift on all azure)

^ plus the corresponding APIs for storage and instance-types

The intent of the PR is to allow a bridge where you can update from here.

dlabrecq: Spoke with @nlcwong and this is expected.

In that case, we don't need that `perspective === DashboardPerspective.ocp` code

Issue #1426, opened 2020-03-25
OpenShift overview page making extra API calls


adberglund:
**Describe the bug**
The OpenShift overview page is still calling for top 3 projects on compute, memory, and storage cards, despite no longer showing this information.

**To Reproduce**
1. Load OpenShift overview page with dev tools open to network tab
2. See extra calls for memory, storage, and compute
Ex: `https://ci.cloud.redhat.com/api/cost-management/v1/reports/openshift/compute/?filter[time_scope_units]=month&filter[time_scope_value]=-1&filter[resolution]=monthly&filter[limit]=3&group_by[project]=*`

**Expected behavior**
We should just be making the call for last month and this month's data

dchorvat1:
verified

PR #1428, opened 2020-03-25
Omit extra API requests for overview

dlabrecq:
Omit extra API requests for overview tabs that are no longer shown

https://github.com/project-koku/koku-ui/issues/1426


dlabrecq:
Thanks, Boaz!

Issue #1446, opened 2020-04-01
Update Infrastructure > OpenShift usage API


dlabrecq:
The "Infrastructure > OpenShift usage" overview page should use the `reports/openshift/costs` API. In addition, we should use the `infrastructure.usage` property for these costs.

dchorvat1:
verified

PR #1447, opened 2020-04-01
Update Infrastructure > OpenShift usage API

dlabrecq:
The "Infrastructure > OpenShift usage" overview page should use the reports/openshift/costs API. In addition, we should use the infrastructure.usage property for these costs.

https://github.com/project-koku/koku-ui/issues/1446


dlabrecq:
Thanks, Boaz!

Issue #1457, opened 2020-04-06
Ocp Usage overview not showing the correct cost


dlabrecq:
The Ocp Usage overview is not showing the correct cost. This page should should zero cost, right now.

For Ocp Usage, we want to use the meta prop `infrastructure.usage.value` as the total cost. However, for the Ocp overview, we still want to use the meta prop `cost.total.value` as the total cost, while using `infrastructure.total.value` for chart data.

dchorvat1:
verified

PR #1458, opened 2020-04-06
Ocp Usage overview not showing the correct cost

dlabrecq:
The Ocp Usage overview is not showing the correct cost. This page should should zero cost, right now.

For Ocp Usage, we want to use the meta prop `infrastructure.usage.value` as the total cost. However, for the Ocp overview, we still want to use the meta prop `cost.total.value` as the total cost, while using `infrastructure.total.value` for chart data.

https://github.com/project-koku/koku-ui/issues/1457


dlabrecq:
Thanks, Boaz!

PR #1479, opened 2020-04-08
Update to PatternFly v4

dlabrecq:
Update to use PatternFly's v4 packages. This includes changes to accommodate PatternFly's breaking release changes.

A couple caveats.

- `Title` components require a `headingLevel` prop, which defaults to `h1`
- `FormGroup` and `TextInput` use the `validated` prop instead of `isValid`

@boaz0 We will need your help to update the Cost model `validated` functionality. This did not work quite as expected. There are 3 states now; default, success, & error states.


dlabrecq:
Will create a draft PR, instead

Issue #1488, opened 2020-04-12
Disable price list filter if there are no rates


boaz0:
**Describe the bug**
When there are no rates in the price list table, disable the filters to not confuse the user with the add rate form.

**To Reproduce**
1. Go to 'Cost models'
2. Click on 'Create cost model'
3. Enter any name and select "Red Hat Openshift Container Platform"
4. Click 'Next'
5. See the error

**Expected behavior**
Disable the filter selectors to imply the user that these are the filters and in order to add a rate she/he needs to click on "Add rate" button.

**Screenshots**
![image](https://user-images.githubusercontent.com/2453279/79067727-2e491180-7cca-11ea-80b6-a992eb49aa74.png)

dchorvat1:
verified

PR #1489, opened 2020-04-12
fix(costmodels): disable price list filter on no rates

boaz0:
related #1488 
image

image


boaz0:
Thanks @dlabrecq

nlcwong:
@boaz0 
Can we make sure that the toolbars are the same for the price list tables?  It looks like we are using compact pagination for the wizard table, but not for the details price list table (which is also throwing off the responsive nature of it.

The empty state message seems to be mis-spaced as well for the price list details table as well - can we get that to be more horizontally centered aligned?

boaz0:
Sure I can work on it @nlcwong

Issue #1497, opened 2020-04-14
Details pages export file naming convention


lcouzens:
## User Story
As a user I would like to see exports from the details pages have a better naming convention on files from exports. For example: app_name-details_name-date so `cost-management-openshift-2020-04-14.csv`

## Impacts
Ex. API, UI, Docs

## Assumptions
-
-

## Acceptance Criteria

- [ ]
- [ ]

dchorvat1:
verified

PR #1507, opened 2020-04-15
Details pages export file naming convention

dlabrecq:
Refactored the details export functionality to better support the file naming convention described below.

>As a user I would like to see exports from the details pages have a better naming convention on files from exports. For example: app_name-details_name-date so cost-management-openshift-2020-04-14.csv

The resulting file names are shown below for the various details pages and group-by combinations.

Screen Shot 2020-04-15 at 5 49 53 PM

FYI, the `js-file-download` code was moved from the store to the details page component, which allows us to customize the filename for each export. In addition, I added the ability to store reports by ID in order to fetch cached reports.

https://github.com/project-koku/koku-ui/issues/1497

### DIFF
--- a/src/api/exports/export.ts
+++ b/src/api/exports/export.ts
@@ -1,0 +1,3 @@
+export interface Export {
+  data: string;
+}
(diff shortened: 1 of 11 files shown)


dlabrecq:
Thanks, @boaz0 !

PR #1517, opened 2020-04-18
fix(costmodels): set a grid breakpoint to the table

boaz0:
image


dlabrecq:
Is there an issue number? I'd like to connect this with an issue in Zenhub so QA can test prior to pushing to prod

dlabrecq:
Created #1520 for QA to track through the Zenhub categories. Please update the issue as necessary.

Issue #1513, opened 2020-04-17
Applied usage date range not displayed correctly when veiwing cluster price list


lcouzens:
**Describe the bug**
Applied usage date range not displayed correctly when viewing cluster price list also cluster per month and node per month metrics show incorrectly.

**To Reproduce**
1. Go to openshift details
2. Click on three dots and `show price list`
3. See pop-up

**Expected behavior**
Text should be displayed correctly

**Screenshots**
![Screenshot from 2020-04-17 12-44-21.png](https://images.zenhubusercontent.com/5b9a6718443d4b342eb0177e/0218c20d-9143-44b8-a2c0-10feddd3824f)

There is also strange brackets after metrics do we need those??

**Additional context**
Testing against PROD, also producible against CI

boaz0:
Can you attach the `/cost-models/<uuid>` response to the issue too?

Thanks.

boaz0:
I am also confused because on `ci-beta` I can see these:

![Screenshot_2020-04-17 Cost Management](https://user-images.githubusercontent.com/2453279/79569059-9d09de80-80bf-11ea-9573-73d1a1c62042.png)

![image](https://user-images.githubusercontent.com/2453279/79569097-ac892780-80bf-11ea-86f7-3325a4e364cd.png)

lcouzens:
Might be related to that :man_shrugging: but cluster/node still seems incorrect for the metric on CI 
![Screenshot from 2020-04-17 13-38-39](https://user-images.githubusercontent.com/18352403/79570188-e6066680-80b0-11ea-98ae-f7fb37860585.png)

boaz0:
Got it, no problem!

PR #1530, opened 2020-04-22
display cluster/node in openshift price list modal

boaz0:
related to #1513

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -413,6 +413,7 @@
       "modal": {
         "applied_usage_date_range": "Applied usage date range",
         "applied_usage_range": "Applied usage range",
+        "cluster currency rate": "Cluster price rate tier {{index}} ({{ unit }})",
         "compute request rate": "Compute request rate tier {{ index }} ({{ unit }})",
         "compute usage rate": "Compute usage rate tier {{ index }} ({{ unit }})",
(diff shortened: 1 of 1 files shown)


src/locales/en.json
```diff
@@ -424,6 +425,7 @@
         "month": "monthly",
         "no_match": "No rates were found",
         "no_range_set": "No range set",
+        "node currency rate": "Node price rate tier {{index}} ({{ unit }})",
```
dlabrecq, line 428: Can we please format keys using the existing naming convention (e.g., "node_currency_rate")?

Also, I don't see "node currency rate" and "cluster currency rate" used anywhere. Is this the only file needing to be updated?

dlabrecq:
Looks like the build requires `syncLocales` to be run.

src/locales/en.json
boaz0: Yes it does come from the API :+1: - otherwise I wouldn't use such a weird naming convention.

dlabrecq: Didn't realize the UI was localizing API strings. 

That will be a problem because if the API changes, it can easily break the UI.

dlabrecq: FYI, I created issue https://github.com/project-koku/koku-ui/issues/1537.

Issue #1504, opened 2020-04-15
Resize charts upon nav toggle event


dlabrecq:
In Cost Management, we set the width of our charts upon a window resize event. However, when the user clicks on the nav toggle, the chart does not resize for the newly available space.

If the user resizes the window in this scenario, the chart will resize as expected. However, when the user clicks on the nav toggle a second time, the chart now overlaps because it's now too wide.

![chrome-capture (2)](https://user-images.githubusercontent.com/17481322/79351011-25448400-7f06-11ea-96dc-46bd5540be4f.gif)

If we could receive an event from the onNavToggle prop of PageHeader; for example, we could resize the chart appropriately when the nav is toggled. Perhaps this could be similar to subscribing to the Insights' `APP_NAVIGATION` event?

See https://projects.engineering.redhat.com/browse/RHCLOUD-5890

dchorvat1:
@dlabrecq 
I'm still seeing the charts overlap the progress bars on CI beta ![Screenshot from 2020-04-28 15-09-11.png](https://images.zenhubusercontent.com/5c3320c5ea2ab97902b1b15b/0d5bdf90-1b3d-442d-99a9-74526962a413)

dlabrecq:
Looks like the timeout needs to be adjusted a bit...

dchorvat1:
verified

PR #1549, opened 2020-04-28
Adjust timeout to resize chart to 500ms

dlabrecq:
Adjusted timeout to resize chart to 500ms. Previous attempt worked better, but could be a little longer for CI beta.

#1504


ddonahue007:
Litmus

PR #1554, opened 2020-05-10
fix(costmodels): disable assigning assigned source to cost model

boaz0:
closes JIRA ticket COST-54

boaz-output


boaz0:
I am going to open a ticket about this on pf-react as I would assume I could disable a checkbox through the Table API.

What I did here is a workaround -> implemented the selection feature by my own.

PR #1556, opened 2020-05-13
fix(costmodel): make it easy to track unassigned sources in table

boaz0:
Remove the "Default cost model" string from all the sources that can be assigned. That way the user can easily see that the ones that are available has no cost model assigned to it and easier to see that those with a cost model attached to it cannot be selected.

Screenshot_2020-05-11 Cost Management(1)


boaz0:
Thanks @ddonahue007

PR #1557, opened 2020-05-13
fix(costmodels): update the assign sources step description

boaz0:
need your review: @dayleparker @nlcwong @lcouzens 

image

Related to https://issues.redhat.com/browse/COST-54

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -329,7 +329,7 @@
     },
     "source": {
       "caption": "Select from the following {{ type }} sources:",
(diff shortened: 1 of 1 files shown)


nlcwong:
@boaz0  Will this PR also solve the issue where we are disabling the user from selecting the row that has a cost model already assigned to it?  If so, then I'm proposing the following wording for this step description:

"Select one or more sources to this cost model. You can skip this step and assign the cost model to a source at a later time. A source will be unavailable for selection if a cost model is already assigned to it."

boaz0:
Done

lcouzens:
Sorry, I took my time getting to this, looks good to me with Natalie's addition :+1:

PR #1570, opened 2020-06-08
fix(costmodel): redirect to cost model after delete

boaz0:


boaz0:
Thanks @ddonahue007 and @dlabrecq ! :heart:

PR #1573, opened 2020-06-11
fix(costmodels): get current cost model from page and not redux

boaz0:


boaz0:
Thank you @dlabrecq for reporting the bug, reviewing and merging the fix and @ddonahue007 for your review!

PR #1586, opened 2020-06-24
refactor(costmodel): clean unnecessary code in cost model header

boaz0:
Remove the `cmpRef` variable - this is no longer needed.
Change the component into a functional component.


boaz0:
Thanks @ddonahue007

PR #1587, opened 2020-06-28
refactor(store): django pagination and query selectors

boaz0:
Small refactor - created a selector for pagination and query data that comes from the Koku backend


boaz0:
Thank you so much @ddonahue007 :bow:

PR #1588, opened 2020-06-29
fix(costmodels): reordered modal buttons correctly

boaz0:
Fix the following modals:
 - assign sources
 - edit markup
 - edit rate
 - add rate


boaz0:
Thank you @dlabrecq and @ddonahue007

PR #1589, opened 2020-06-30
fix(costmodels): show close dialog on wizard close if >20% is filled

boaz0:
Display the exit confirmation dialog in the following cases:
 - Cost model type is Amazon or AWS and the user is in one of these steps: sources, review
 - Cost model type is Openshift and the user is in one of these steps: markup, sources, review
 - Cost model type is Openshift and the user is in the price list step and added at least one rate


nlcwong:
LGTM

boaz0:
Thanks you @nlcwong and @ddonahue007 :bow:

PR #1592, opened 2020-07-02
fix(costmodels): use compact pagination + remove view opt from kebab

boaz0:
image

1) the delete option in the kebab option is black
2) the view cost model was removed from the kebab option
3) the upper pagination is in compact variation


boaz0:
Thanks @ddonahue007 :pray:

PR #1599, opened 2020-07-05
fix(costmodels): update cost model details header title size + padding

boaz0:
image


nlcwong:
The padding still looks a bit off.  Like there is too much at the top of the table column header (should be medium at 16px?).  As well as the padding above the toolbar. I believe the margin all around the table should be large (24px).

boaz0:
image

boaz0:
image

PR #1600, opened 2020-07-06
fix(costmodels): remove padding, add pagintn footer, delete color black

boaz0:
image

@nlcwong can you take a look?


boaz0:
Thank you @ddonahue007

PR #1602, opened 2020-07-07
fix(costmodels): remove name from title in the cost model edit dialog

boaz0:
image


boaz0:
Thanks everybody! @dlabrecq @ddonahue007 :bow:

PR #1611, opened 2020-07-12
fix(costmodels): Fix pagination settings in the sources tab

boaz0:
- Make the pagination on the toolbar compact
- Align the pagination on bottom to right


boaz0:
Thanks @ddonahue007 :cactus:

PR #1612, opened 2020-07-13
fix(costmodels): fix redirect issue on cost model update

boaz0:
This fixes a bug when updating a cost model.
The problem was due to the changes I did to the cost model routing and now that cost models details page and cost model details page are two different pages, I had to update the Redux action as well.

Now, when the cost model is successfully updated we will fetch it specifically and not all the cost models.

//cc @lcouzens


boaz0:
Thank you @ddonahue007 & @dlabrecq

PR #1615, opened 2020-07-15
fix(costmodels): edit markup dialog updates

boaz0:
This change will make sure that the precision of the markup is not more than 2 digits after the dot.

//cc @nlcwong

### DIFF
--- a/src/pages/costModels/costModelsDetails/components/updateMarkupDialog.tsx
+++ b/src/pages/costModels/costModelsDetails/components/updateMarkupDialog.tsx
@@ -6,7 +6,13 @@ import {
   InputGroup,
   InputGroupText,
   Modal,
+  Stack,
+  StackItem,
+  Text,
+  TextContent,
   TextInput,
+  TextVariants,
+  Title,
 } from '@patternfly/react-core';
 import { CostModel } from 'api/costModels';
 import React from 'react';
@@ -14,6 +20,7 @@ import { InjectedTranslateProps, translate } from 'react-i18next';
 import { connect } from 'react-redux';
 import { createMapStateToProps } from 'store/common';
 import { costModelsActions, costModelsSelectors } from 'store/costModels';
+import { formatValue } from 'utils/formatValue';
 
 interface Props extends InjectedTranslateProps {
   isLoading: boolean;
@@ -31,7 +38,7 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
   constructor(props) {
     super(props);
     this.state = {
-      markup: String(this.props.current.markup.value || 0),
+      markup: formatValue(Number(this.props.current.markup.value), 'markup', {fractionDigits: 2}) as string ||  '0.00',
     };
   }
   public render() {
@@ -45,9 +52,7 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
     } = this.props;
     return (
       <Modal
-        title={t('cost_models_details.edit_markup', {
-          cost_model: current.name,
-        })}
+        title={t('cost_models_details.edit_markup')}
         isOpen
         onClose={() => onClose({ name: 'updateMarkup', isOpen: false })}
         variant="small"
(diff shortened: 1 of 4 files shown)


src/pages/costModels/components/logic/markup.ts
```diff
@@ -0,0 +1,11 @@
+export const hasPrecisionOf = (afterDot: number) => (
+  markup: string
+): boolean => {
+  if (markup.split('.').length === 2) {
+    const digits = markup.split('.')[1];
+    if (digits.length > afterDot && !isNaN(Number(digits))) {
+      return true;
+    }
+  }
```
dlabrecq, line 9: Instead of manually splitting, we have some formatting utils in src/utils/formatValue.ts.

In particular, I'm thinking `formatCurrency` would work here. You can specify the number of `fractionDigits` and the currency can be localized (eventually) thru this function.

boaz0: Sorry, I was trying to use `formatCurrency` and `unknownTypeFormat` and it didn't work.
Lets say the user enter `0` in the markup then both of them will return `0.00`. Now how am I supposed to compare between two to know whether to update or not to update.

dlabrecq: I was only referring to `formatCurrency` as an example. We have other utils to format usage hours, GB, etc. 

My thought was that you could create another function for percentage? For example, you could set the `minimumFractionDigits` to 0 and `maximumFractionDigits` as 2. 

The `unknownTypeFormatter` function sets both equally, so that's why you always end up with 0.00.

dlabrecq: I was referring to `formatCurrency` as an example. We have other utils to format usage hours, GB, etc. 

My thought was that you could create another function for percentage? For example, you could set the `minimumFractionDigits` to 0 and `maximumFractionDigits` as 2. 

The `unknownTypeFormatter` function sets both equally, so that's why you always end up with 0.00.

We will need to localize percentages, along with currency, so hoping this or a similar function will help? I know the backend folks are starting work to localize currency soon.

https://stackoverflow.com/questions/49439198/localizing-percentage-values

boaz0: @dlabrecq I still cannot realy on `toLocaleString` to determine that.
What happens when the user enters 10000 then it will return `100,00` or at some big numbers the numbers after the digits are discarded.
Sorry I cannot see how it is going to help me.

By the way, this is my code:

```
# utils/formatValue.ts

export const formatPercentages: ValueFormatter = (
  value,
  _unit,
  { fractionDigits = 2, minimumFractionDigits = 0 } = {}
) => {
  let fValue = value;
  if (!value) {
    fValue = 0;
  }
  return fValue.toLocaleString('en', {
    minimumFractionDigits: minimumFractionDigits,
    maximumFractionDigits: fractionDigits,
  });
};

# pages/costModels/components/logic/markup.ts

import { formatPercentages } from 'utils/formatValue';

export const hasPrecisionOf = (prev: string, markup: string): string => {
  const value = Number(markup)
  if(isNaN(Number(prev))) {
    if(isNaN(value)) {
      return markup
    }
    return formatPercentages(value) as string;
  }
  const twodigits = formatPercentages(value) as string
  const threedigits = formatPercentages(value, '', {minimumFractionDigits: 0, fractionDigits: 3}) as string
  if (twodigits !== threedigits) {
    return twodigits as string
  }
  return markup
};

# pages/costModels/createCostModelWizard

....
          onMarkupChange: (value) => {
            this.setState({markup: hasPrecisionOf(this.state.markup, value)})
          },
...
```

This still has edge cases.

dlabrecq: Perhaps I don't understand the edge cases; `twodigits` Vs `threedigits`, returning `isNaN` markup, etc. I assumed valid percentages are numbers and use the same decimal precision?

It's odd that `toLocaleString` would drop numbers because we use it for large numbers throughout the UI.

Don't want to hold up the PR, but we will need to localize eventually.

dlabrecq: Perhaps I don't understand the edge cases; twodigits Vs threedigits, returning isNaN markup, etc. I assumed valid percentages are numbers and use the same decimal precision?

It's odd that toLocaleString would drop digits because we use it for large numbers throughout the UI.

Don't want to hold up the PR, but we will ultimately need to work with localized values. Decimals, commas, percent symbol, etc. appear differently per locale, so markup.split('.') will likely fail.

If not with toLocaleString, then perhaps we can use Intl.NumberFormat? I suspect even (10.8).toFixed(2) may be better than using split?

https://stackoverflow.com/questions/1726630/formatting-a-number-with-exactly-two-decimals-in-javascript

boaz0:
@dlabrecq can you please review this again? I hope I addressed all your concerns.

dlabrecq:
@boaz0 Please hold off merging until PRs have been reviewed

src/pages/costModels/createCostModelWizard/index.tsx
```diff
@@ -223,7 +224,15 @@ class CostModelWizardBase extends React.Component<Props, State> {
...
-          onMarkupChange: value => this.setState({ markup: value }),
+          onMarkupChange: value => {
+            const markupDecimal = Number(value)
+            const dx = value.split('').findIndex(c => c === '.')
+            if(!isNaN(markupDecimal) && dx > -1 && value.length - dx - 1 > 2) {
+              this.setState({markup: formatValue(markupDecimal, 'markup', {fractionDigits: 2}) as string})
+              return
+            }
+            this.setState({ markup: value });
+          },
```
dlabrecq, line 235: Not clear as to what changed, but this code still appears to split a string? Not all locales will use a decimal

src/pages/costModels/costModelsDetails/components/updateMarkupDialog.tsx
```diff
@@ -88,35 +93,57 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
...
+                    aria-label={t('cost_models_wizard.markup.markup_label')}
+                    id="markup-input-box"
+                    value={this.state.markup}
+                    onChange={(markup: string) => {
+                      const markupDecimal = Number(markup)
+                      const dx = markup.split('').findIndex(c => c === '.')
+                      if(!isNaN(markupDecimal) && dx > -1 && markup.length - dx - 1 > 2) {
+                        this.setState({markup: formatValue(markupDecimal, 'markup', {fractionDigits: 2}) as string})
+                        return
+                      }
```
dlabrecq, line 134: This code also appears to split a string and search for a decimal character.

PR #1618, opened 2020-07-16
COST-134: Tag list modal consistency update

dlabrecq:
Updated tags modal shown on details page per new mocks. However, we're using PatternFly's data list component instead of a simple HTML table -- reviewed with Natalie.

https://issues.redhat.com/browse/COST-134

Screen Shot 2020-07-16 at 3 29 18 PM

### DIFF
--- a/src/pages/details/components/tag/tagLink.tsx
+++ b/src/pages/details/components/tag/tagLink.tsx
@@ -77,19 +77,19 @@ class TagLinkBase extends React.Component<TagLinkProps> {
     const { filterBy, groupBy, id, report, reportPathsType } = this.props;
     const { isOpen } = this.state;
 
-    const tagKeyValues = [];
+    let tagKeyValuesCount = 0;
 
     if (report) {
       for (const tag of report.data) {
-        for (const val of tag.values) {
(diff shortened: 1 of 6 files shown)


src/pages/details/components/tag/tagLink.tsx
```diff
@@ -77,19 +77,19 @@ class TagLinkBase extends React.Component<TagLinkProps> {
...
     const { isOpen } = this.state;
 
-    const tagKeyValues = [];
+    let tagKeyValuesCount = 0;
 
     if (report) {
       for (const tag of report.data) {
-        for (const val of tag.values) {
-          tagKeyValues.push(`${(tag as any).key}: ${val}`);
+        for (const {} of tag.values) {
```
boaz0, line 84: why not doing `tagKeyValuesCount += tag.values.length()`

dlabrecq: Yes, that would be better. This was old code and didn't see that. Thanks!

dlabrecq: Yes, that would be better. This was an old code block, previously used to concatenate strings. Thanks!

PR #1619, opened 2020-07-16
COST-342: Azure breakdown page fix

dlabrecq:
This fixes an issue with the Azure breakdown page, due to the AWS OU refactor. While the page is loading, the `query` object may be available yet; thus, we must check for undefined `query` objects.

https://issues.redhat.com/browse/COST-342


src/pages/details/awsBreakdown/awsBreakdown.tsx
```diff
@@ -61,9 +61,9 @@ const mapStateToProps = createMapStateToProps<
       time_scope_value: -1,
       resolution: 'monthly',
       limit: 3,
-      ...(query.filter.account && {['account']: query.filter.account})
+      ...(query && query.filter && query.filter.account && {['account']: query.filter.account})
```
boaz0, line 64: nitpick: I believe `query && query.filter_by` will do the same. But if ternary operation is more readable to you then keep it.

dlabrecq: The problem was that `query` can be undefined -- the code dies here without. First need to check if `query` exists, then check if `query.filter` exists.

PR #1621, opened 2020-07-21
fix(costmodels): assign-sources/pricelist fit & finish

boaz0:
fixes COST 368


boaz0:
Thank you @ddonahue007

PR #1623, opened 2020-07-23
COST-374: Cost model empty state appears at bottom of page

dlabrecq:
Fixed cost models empty state so it's not at the bottom of the page

https://issues.redhat.com/browse/COST-374

**Before**
Screen Shot 2020-07-23 at 3 32 43 PM

**After**
Screen Shot 2020-07-23 at 6 22 36 PM


nlcwong:
LGTM

PR #1628, opened 2020-08-02
chore(lint): fix tslint/prettier issue

boaz0:
This is supposed to fix formatting problems we see in here: https://github.com/project-koku/koku-ui/pull/1627#discussion_r462634288

What was the problem?
First, `prettier` was removed from `package.json` and therefore, if you remove `node_modules` and run `hasky` the command will fail even though everything is OK.
In addition, `tslint` by itself does not fix the errors and requires `-p` to be specified with the path to `tsconfig.json`.
Also, while working on this removed `js,jsx` from the files since we are using only typescript and added locales linting.

## So why did it fail on my machine?
My assumption is that `prettier` was not removed from `node_modules/.bin`
## Then why it formatted your code that way?
I had that issue because I was running `koku-ui` on a different machine. I saw that error and installed `prettier` manually which caused it not to be installed on `node_modules/.bin` but on `/usr/local/` and ignored the `prettierrc` file.


package.json
```diff
@@ -127,7 +122,9 @@
     "jest-emotion": "10.0.32",
     "lint-staged": "7.2.0",
     "ts-jest": "26.1.1",
-    "webpack-dev-server": "^3.11.0"
+    "webpack-dev-server": "^3.11.0",
+    "tslint": "5.10.0",
+    "tslint-config-prettier": "1.13.0"
```
dlabrecq, line 127: Please assign in alphabetical order

tslint.json
```diff
@@ -18,6 +18,7 @@
     "jsx-wrap-multiline": true,
     "object-literal-sort-keys": false,
     "quotemark": [true, "single", "jsx-double"],
+    "arrow-parens": [true, "ban-single-arg-parens"],
```
dlabrecq, line 21: Can we remove the arrow parens? I don't believe PatternFly requires them -- I'd like to follow their lint / prettier rules if possible?

Specifically, I'd like us to use the `printWidth` prop in `.prettierrc`


```
{
  "semi": true,
  "singleQuote": true,
  "tabWidth": 2,
  "trailingComma": "none",
  "useTabs": false,
  "printWidth": 120
}
```

dlabrecq: Thanks for digging into this, BTW. :)

package.json
```diff
@@ -29,7 +29,6 @@
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": [
       "yarn lint:precommit --fix",
```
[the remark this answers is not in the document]
boaz0: But you need to run eslint/prettier - that is whay lint:precommit does, right? So removing it will not run eslint/prettier.

package.json
```diff
@@ -28,8 +26,7 @@
   },
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": [
-      "yarn lint:precommit --fix",
-      "prettier --write",
+      "yarn lint:ts --fix",
       "git add"
     ]
   },
```
dlabrecq, line 32: This appears to be unused, so you can remove it

package.json
```diff
@@ -117,15 +111,24 @@
...
     "enzyme": "3.8.0",
     "enzyme-adapter-react-16": "1.7.1",
     "enzyme-to-json": "3.3.5",
+    "eslint": "5.6.0",
+    "eslint-plugin-jsdoc": "^21.0.0",
+    "eslint-plugin-markdown": "^1.0.1",
+    "eslint-plugin-patternfly-react": "^4.1.2",
+    "eslint-plugin-prettier": "^3.1.2",
+    "eslint-plugin-react": "^7.18.0",
     "husky": "0.14.3",
```
dlabrecq, line 127: Husky appears to be unused.

package.json
```diff
@@ -117,15 +111,24 @@
...
     "jest": "26.1.0",
     "jest-emotion": "10.0.32",
     "lint-staged": "7.2.0",
```
dlabrecq, line 130: Looks like we can remove the `lint-staged` dependency as well

package.json
boaz0: done

package.json
boaz0: done

package.json
boaz0: done

PR #1634, opened 2020-08-07
PatternFly milestone update

dlabrecq:
Updated @patternFly and r@edhat-cloud-services component packages to latest milestones.

In addition, I've refactored all @patternfly/react-token (and a couple remaining @patternfly/react-icons) imports to use full paths. This cuts our build time down by more than half; from 4 min 35 sec to approx 1 min 50 sec.

https://issues.redhat.com/browse/COST-416


src/pages/costModels/costModel/index.tsx
```diff
@@ -6,7 +6,7 @@ import {
   TabContent,
   Title,
 } from '@patternfly/react-core';
-import { ErrorCircleOIcon } from '@patternfly/react-icons';
+import { ErrorCircleOIcon } from '@patternfly/react-icons/dist/js/icons/error-circle-o-icon';
```
boaz0, line 9: why are we importing this from 'dist/js' and not from '@patternfly/react-icons'?

src/pages/costModels/costModelsDetails/components/markup.styles.ts
```diff
@@ -1,4 +1,5 @@
-import { global_FontSize_xl, global_spacer_md } from '@patternfly/react-tokens';
+import global_FontSize_xl from '@patternfly/react-tokens/dist/js/global_FontSize_xl';
```
boaz0, line 1: Also here, is this how we suppose to import?

src/pages/costModels/costModel/index.tsx
dlabrecq: By including exactly what we need, webpack doesn't need to filter thousands of tokens and icons. Webpack opens and parses thru patternfly's 3k+ files to find what we need. Each time we do that, it adds .5-1 sec to the build.

src/pages/costModels/costModelsDetails/components/markup.styles.ts
dlabrecq: Yes. Please see the description above

boaz0:
LGTM

PR #1642, opened 2020-08-30
refactor(costmodels): make cost models header and table composable

boaz0:
reduce props-drilling and code duplication


src/pages/costModels/components/paginationToolbarTemplate.tsx
```diff
@@ -0,0 +1,56 @@
...
+  t: TranslationFunction,
+  props: PaginationToolbarTemplateProps
+) => {
+  return {
+    ...props,
+    'aria-label': t(props['aria-label']),
+  };
+};
+
+export default withI18nProps(translatePaginationToolbarProps)(
```
dlabrecq, line 54: It's not clear why `TranslationFunction` is used. Can we not use `translate()` here?

dlabrecq (commented):
I'm all for reducing code duplication. Although, the template and translatable props look quite complex IMHO. My only concern is that this may be more difficult for other folks to read and maintain in the future?

src/pages/costModels/components/paginationToolbarTemplate.tsx
boaz0: I don't like `translate()` because it adds the `t` functions to my props which makes components coupled to it. In addition, running tests on the `translate` components fails or if I am wrapping it with the I18n provider it just does not render. However, if I am using the render-props version:
```
return (
  <I18n>{t => { ... }</I18n>
)
```

the tests does not fail and everything works fine.

That being said, I can discard `withI18nProps` in favor of render props. Will that make it clear?

dlabrecq: That's ok. Thanks for the explanation.

boaz0: Thanks. I updated the PR and discarded the `withI18nProps`.

dlabrecq (approved):

boaz0:
Thanks @dlabrecq and @ddonahue007

PR #1649, opened 2020-09-05
Inactive sources cookie

dlabrecq:
This sets a session cookie to ensure the inactive sources alert will remain closed while the user is logged in. That is, providing the user closed the inline alert.

If the user logs out or closes the browser, the session cookie will be invalid and removed upon a new login. In this scenario, the alert will be shown, again.

If there are no longer inactive sources, the cookie is also removed. Thus, allowing new alerts to be shown.

https://issues.redhat.com/browse/COST-463

demo2


boaz0 (approved):

ddonahue007 (commented):
LGTM

PR #1651, opened 2020-09-06
feat(costmodels): add tagging base rates

boaz0:
Closes JIRA ticket COST-79

- [x] https://github.com/project-koku/koku/pull/2375 is merged
- [x] get ~100% or at least~ 90% for diff coverage
- [x] https://github.com/project-koku/koku/pull/2395 is merged
## wizard - form
image
## wizard - table
image
## add rate modal
image
image

## cost model - edit rate modal
image
## cost model - price list table
image


nlcwong:
1. Wizard
- Step 2 should be "Create price list"
2. Wizard table
- The table w/in the expansion panel should not have row separator lines showing, its supposed to be a compact table which I believe does not include the row lines in between rows.
- The rates for tags need to be formatted the same way as rates in the main table ($xx.xx) where its 2 decimal places are always filled, even if its .00.
3. Add rate modal
- I didn't think we allowed users to enter a fraction of a cent (e.g. $0.001), @adberglund do we know if we want to restrict users from entering fractions of a cent for the rates?
4. cost model - edit rate modal
- Make sure that we only enable the "Save" button when we know contents have been changed (could not tell from the screenshot if you had already edited the contents)
5. cost model - price list table
- Same comment that the tag rates table should be a compact table that has small row heights and no row separator lines in between the rows.
- Tag rates need to follow same currency formatting as the regular rates.

boaz0:
@nlcwong 

1) Wizard: Step 2 should be "Create price list"
2) The table w/in the expansion panel should not have row separator lines showing, its supposed to be a compact table which I believe does not include the row lines in between rows.
3) The rates for tags need to be formatted the same way as rates in the main table ($xx.xx) where its 2 decimal places are always filled, even if its .00.

image

4) edit rate modal: make sure that we only enable the "Save" button when we know contents have been changed (could not tell from the screenshot if you had already edited the contents)

image
image
image

image
image
image
image
image
image

5) cost model - price list table
image

nlcwong:
LGTM

dlabrecq (approved):
Looks good to me.

Might want to confirm that the `I18N` components revceive the correct locale? 

In App.tsx, we provide the locale to `I18nProvider`. You could change that locale to "fr", for example.

src/pages/costModels/components/rateTable.test.tsx
```diff
@@ -0,0 +1,72 @@
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { Rate } from '../../../api/rates';
```
dlabrecq, line 4: If you could, please replace relative paths with absolute paths

boaz0: Thanks. I fixed that.
In addition, I checked that `I18n` uses the correct locale. Although I had to also set `fallback` to `fr` in the `I18nProvider.tsx`.

boaz0:
Thanks @dlabrecq and @nlcwong

PR #1658, opened 2020-09-11
createMapStateToProps requires props

dlabrecq:
It appears that `createMapStateToProps` requires `props` in order to update the page properly. Without, users cannot update the details table with a new group by selection.


ddonahue007 (commented):
LGTM

boaz0:
LGTM

PR #1660, opened 2020-09-15
Fixed error message so that it matches target name.

ddonahue007:
Fixed error message so that it matched target name.


dlabrecq (approved):

boaz0:
LGTM

PR #1661, opened 2020-09-15
AWS details not displaying OU names on filtered data

dlabrecq:
The API returns different property names if we're filtering AWS org units. This ensures the ID is always used as a fallback.

https://issues.redhat.com/browse/COST-524

Screen Shot 2020-09-15 at 2 39 06 PM


ddonahue007 (approved):
LGTM

boaz0 (commented):
LGTM

PR #1675, opened 2020-09-28
Export fails when doing view by OU and filter by OU

dlabrecq:
The CSV export modal is generating a query which is not supported by the API.

"Multiple org_unit_id must be represented with the or: prefix."

https://issues.redhat.com/browse/COST-560


ddonahue007 (approved):
LGTM

boaz0 (approved):

PR #1679, opened 2020-10-01
Remove "emotion" from aws details table

ddonahue007:
Effected Directory:  src/pages/awsDetails

Screen Shot 2020-10-01 at 9 54 33 AM


src/pages/details/awsDetails/detailsTable.styles.ts
```diff
@@ -1,73 +0,0 @@
...
-    position: 'relative',
-  },
-  infoArrowDesc: {
-    bottom: global_spacer_xs.value,
-  },
-  infoDescription: {
-    color: global_disabled_color_100.value,
-    fontSize: global_FontSize_xs.value,
-  },
-} as { [className: string]: React.CSSProperties };
```
dlabrecq, line 30: We don't need to convert these properties, since it does not use Emotion. Only the styles wrapped with `css` need to be converted. It can be implemented either way, but just more to test visually.

The Azure and OpenShift tables will also have a similar "styles" implementation. For consistency, it may be best to revert this one?

dlabrecq (changes requested):

src/pages/details/awsDetails/detailsTable.styles.ts
ddonahue007: I can revert these changes, I know when we talked that we  were just interested in changing over the emotion stuff. In this case I thought it was going to be cleaner not to have both files.

dlabrecq: Considering the we had a couple CSS bugs that needed to be fixed, I would prefer to limit the changes to just Emotion for now. This would help eliminate any surprises in production.

dlabrecq (approved):
LGTM, thanks!

PR #1683, opened 2020-10-04
Invalid workflow file

dlabrecq:
Github encountered an error with the codeQL workflow file.

"you may only define one of `paths` and `paths-ignore` for a single event"

https://github.com/project-koku/koku-ui/actions/runs/287204312


boaz0 (commented):
LGTM

PR #1688, opened 2020-10-05
Allow rate or cost_model resource

adberglund:
## Summary
We have renamed the rbac resource from rate to cost_mode. This will allow us to respect the new value and catch any existing rbac roles with rate in use instead.

### DIFF
--- a/src/store/rbac/selectors.ts
+++ b/src/store/rbac/selectors.ts
@@ -20,7 +20,7 @@ export const isCostModelWritePermission = (state: RootState) => {
   if (app === 'cost-management' && resource === '*' && operation === '*') {
     return true;
   }
-  if (resource === 'rate' && operation === 'write') {
+  if ((resource === 'rate' || resource === 'cost_model') && operation === 'write') {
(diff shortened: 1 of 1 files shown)


src/store/rbac/selectors.ts
```diff
@@ -20,7 +20,7 @@ export const isCostModelWritePermission = (state: RootState) => {
   if (app === 'cost-management' && resource === '*' && operation === '*') {
     return true;
   }
-  if (resource === 'rate' && operation === 'write') {
+  if ((resource === 'rate' || resource === 'cost_model') && operation === 'write') {
```
boaz0, line 23: Are you keeping `rate` for backward compatibility, aren't you?

boaz0 (approved):

ddonahue007 (approved):
LGTM

src/store/rbac/selectors.ts
adberglund: @boaz0 Yes.

PR #1694, opened 2020-10-09
Updated pkgs based on yarn audit

dlabrecq:
Updated pkgs based on yarn audit.

- `axios` had a moderate level vulnerability.
- `i18next-express-middleware` is unused.
- `i18next-json-sync` still has a low level vulnerability

Note that `i18next-json-sync` is used, during build time only, to check if localization files are in sync. It's already at its latest version,  but we can at least make this as a dev-dependency in package.json.

https://issues.redhat.com/browse/COST-514


ddonahue007 (approved):
LGTM

boaz0:
LGTM

PR #1695, opened 2020-10-12
Updated PatternFly to install the latest Victory packages

dlabrecq:
Updated PatternFly to install the latest Victory packages. 

Note that this does not solve the issue below, but the first step is to use the latest Victory chart packages.

https://issues.redhat.com/browse/COST-523


ddonahue007 (approved):
LGTM

PR #1696, opened 2020-10-13
UI overview tooltip not working correctly

dlabrecq:
When charts are re-rendered, the custom legend tooltip stops working. This happens whenever the user changes tabs or resizes the browser window.

The fix is to simply re-order how the "voronio" and "cursor" container properties are merged via Victory's `createContainer` function. Thus, saving the container state is no longer necessary.

https://issues.redhat.com/browse/COST-523

chrome-capture


ddonahue007 (approved):
LGTM

PR #1697, opened 2020-10-13
Historical chart scale is too small

dlabrecq:
The scale (aspect ratio) is off for historical charts. This appears to have changed with the latest Victory packages.

This change ensures each chart container defines both height and width.

https://issues.redhat.com/browse/COST-626

**Before**

Screen Shot 2020-10-12 at 10 37 19 PM

**After**

Screen Shot 2020-10-12 at 10 55 46 PM


ddonahue007 (approved):
LGTM

PR #1698, opened 2020-10-13
Show loading state while fetching details table data

dlabrecq:
For the AWS/Azure/Ocp details page, a loading (spinning circle) is shown inside the table body. However, it's so brief that it looks like an empty table. It would be preferable to continue showing the loading state used while fetching providers.

https://issues.redhat.com/browse/COST-627


ddonahue007 (approved):
LGTM

PR #1707, opened 2020-10-19
Last of the emotion removals

ddonahue007:
Remove emotion form the overView
This is the last of the emotion usages so remove emotion dependencies as well

Screen Shot 2020-10-19 at 4 35 41 PM


dlabrecq (approved):
LGTM. Thanks for all the work to remove Emotion!

PR #1709, opened 2020-10-23
Use RBAC permissions to provide user access to pages

dlabrecq:
As a user, I expect UI pages to be shown according to my user's entitlements and RBAC permissions. Pages should only be shown if the following conditions are met.

1. The user has entitlements
2. If the user is an org admin, all pages are available
3. A page is shown if the user has RBAC permissions for specific resource types

As an example, consider the `cost-demo` user. Since that user is an org admin, they should have access to all pages.

The `cost-user` user is not an org admin and only has the `cost_model` resource type defined as read/write permissions. In this scenario, the user only has access to the cost models page. 

When users don't have permissions, they see the `NotAuthorized` empty state.

For more details, see https://issues.redhat.com/browse/COST-646

**The `cost-user` user and the overview page**
Screen Shot 2020-10-22 at 11 34 43 PM

**The `cost-user` user and the cost models page**
Screen Shot 2020-10-22 at 11 34 58 PM

### DIFF
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -3,9 +3,9 @@ import { getProvidersQuery } from 'api/queries/providersQuery';
 import { AxiosError } from 'axios';
 import { I18nProvider } from 'components/i18n';
 import { InactiveSources } from 'components/sources/InactiveSources/InactiveSources';
-import Maintenance from 'pages/state/maintenance/maintenance';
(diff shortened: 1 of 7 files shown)


ddonahue007 (commented):
LGTM

boaz0 (commented):
LGTM :+1:

PR #1711, opened 2020-10-23
Breakdown chart shows markup and raw cost values switched

dlabrecq:
The markup and raw cost values shown in the breakdown chart's legend are switched.

https://issues.redhat.com/browse/COST-640


ddonahue007 (approved):
LGTM

PR #1712, opened 2020-10-23
Moved InactiveSources

dlabrecq:
Moved `InactiveSources` to the `PermissionsComponent` wrapper.

This simplifies `App.tsx` and ensures `InactiveSources` is not rendered along with `NotAuthorized`. Thus, the sources (provider) API requests are never generated unless the user has RBAC permissions.

https://issues.redhat.com/browse/COST-646


ddonahue007 (approved):
LGTM

PR #1713, opened 2020-10-23
Use RBAC permissions to show/hide tabs and perspective options

dlabrecq:
As a user, I expect Overview tabs and perspective options to be shown according to my user's entitlements and RBAC permissions. Items should only be shown if the following conditions are met.

1. If the user is an org admin, all items are available
2. An item is shown if the user has RBAC permissions for specific resource types
3. Sources are populated with data

https://issues.redhat.com/browse/COST-647


ddonahue007 (approved):
LGTM

PR #1717, opened 2020-10-26
chore(routes): add additional app data to improve Pendo

boaz0:
closes COST 654

### DIFF
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -32,10 +32,11 @@ export class App extends React.Component<AppProps, AppState> {
   public state: AppState = { locale: 'en', maintenanceMode: false };
 
   public componentDidMount() {
-    const { history } = this.props;
+    const { history, location } = this.props;
 
     insights.chrome.init();
     insights.chrome.identifyApp('cost-management');
+    insights.chrome.appAction(location.pathname);
 
     this.appNav = insights.chrome.on('APP_NAVIGATION', event => {
       const currRoute = routes.find(({ path }) => path.includes(event.navId));
@@ -50,11 +51,13 @@ export class App extends React.Component<AppProps, AppState> {
 
     if (location && location.pathname !== prevProps.location.pathname) {
       window.scrollTo(0, 0);
+      insights.chrome.appAction(location.pathname);
     }
   }
 
   public componentWillUnmount() {
     this.appNav();
+    insights.chrome.appAction(undefined);
   }
 
   public render() {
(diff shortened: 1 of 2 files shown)


src/pages/costModels/costModel/index.tsx
```diff
@@ -47,7 +47,24 @@ class CostModelInformation extends React.Component<Props, State> {
...
+  }
+
+  public componentWillUnmount() {
+    insights.chrome.appObjectId(undefined);
+  }
+
   public componentDidMount() {
+    const { uuid } = this.props.match.params;
+    const { tabIndex } = this.state;
+    insights.chrome.appObjectId(`uuid:${uuid}_tabIndex:${tabIndex}`);
```
dlabrecq, line 67: Is identifying each table item with a UUID necessary? Do these particular UUIDs (e.g., a9ffd77a-b8ea-4107-8836-78ad618fb2a8) convey any useful / readable information?

Pendo will be notified when the cost models and/or cost models details pages are loader, but probably don't need to identify the individual table item. It's still the same details page, just viewed with different data.

I'm thinking that we could provide a simpler ID for each page? For example, 'cost-models', 'cost-models-details', etc. The tab index may be useful, but suspect the UUIDs are a bit much.

We could reach out to a designer and ask exactly what are they looking for from the cost models details page? However, designers don't typically ask for this level of detail in their "design share" meetings.

That said, the `componentDidMount` code could be located in a component wrapper (i.e., if we use our page paths as a generic `appObjectId`)?

dlabrecq (changes requested):

src/pages/costModels/costModel/index.tsx
boaz0: Um, good question. I think it can be useful but not necessary.

src/utils/pendoRouter.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const PendoRoute: React.FunctionComponent<RouteProps & PendoProps> = props => {
+  const { labelKey, ...routeProps } = props;
+  React.useEffect(() => {
+    const labelKeySliced = labelKey.split('.');
+    const pageName = labelKeySliced[labelKeySliced.length - 1];
+    insights.chrome.appAction(pageName);
+    return () => insights.chrome.appAction(undefined);
+  });
+  return <Route {...routeProps} />;
```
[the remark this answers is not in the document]
boaz0: Yea, that's a good idea!

src/pages/costModels/costModel/index.tsx
dlabrecq: Don't feel strongly about it, just curious.

dlabrecq (approved):

dlabrecq:
Just need to check `location` is not undefined for the tests to pass (e.g., `location & location.pathname`)

PR #1720, opened 2020-10-28
fix(costmodels): add aria-label to Modal with custom header

boaz0:
closes COST 656

### DIFF
--- a/src/pages/costModels/createCostModelWizard/index.tsx
+++ b/src/pages/costModels/createCostModelWizard/index.tsx
@@ -336,6 +336,7 @@ class CostModelWizardBase extends React.Component<Props, State> {
           }}
         />
         <Modal
+          aria-label={t('cost_models_wizard.confirm.title')}
           isOpen={this.state.isDialogOpen}
           header={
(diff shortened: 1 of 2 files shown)


src/pages/costModels/createCostModelWizard/index.tsx
```diff
@@ -336,6 +336,7 @@ class CostModelWizardBase extends React.Component<Props, State> {
           }}
         />
         <Modal
+          aria-label="quit-wizard-dialog"
```
dlabrecq, line 339: The aria labels should be localized, so it can be read by a screen reader

dlabrecq (changes requested):

src/pages/costModels/createCostModelWizard/index.tsx
boaz0: Oh, I didn't know that. No problem. I will fix that.

dlabrecq (approved):

boaz0:
Thanks @dlabrecq

PR #1727, opened 2020-11-10
Rm dead i18n

ddonahue007:
Cleanup of i18n keys, first of many. I am keeping these PR's small to make reviews easier

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -1266,9 +1266,7 @@
     "gb": "{{value}} GB",
     "gb-hours": "{{value}} GB-hours",
     "gb-mo": "{{value}} GB-month",
-    "hours": "{{value}} hours",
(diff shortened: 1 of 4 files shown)


src/locales/en.json
```diff
@@ -1266,9 +1266,7 @@
     "gb": "{{value}} GB",
     "gb-hours": "{{value}} GB-hours",
     "gb-mo": "{{value}} GB-month",
-    "hours": "{{value}} hours",
     "hrs": "{{value}} hours",
-    "usd": "{{value}}",
```
dlabrecq, line 1271: I believe 'usd' is a default value, since there is no string. Let's keep it.

dlabrecq (changes requested):

dlabrecq (approved):

PR #1728, opened 2020-11-11
Rm dead i18n

ddonahue007:
Cleanup of i18n keys that are no longer being used.

TOTALS:
FOUND                         505
NOT FOUND                360
EXCLUDED                   19


dlabrecq (approved):
LGTM

PR #1730, opened 2020-11-12
Wrong month label shown on breakdown page

dlabrecq:
We're currently showing the previous month instead of the current month in the breakdown page header. 

The problem is that default month offset should not be used here to display the current month.

https://issues.redhat.com/browse/COST-725

Screen Shot 2020-11-12 at 12 31 21 PM


ddonahue007 (approved):
LGTM

PR #1732, opened 2020-11-12
Fixed tag names being used in Azure widgets

ddonahue007:
We had a AWS i18n tag being used in the Azure widgets. This was not causing any issues as the key values are the same.


dlabrecq (approved):
Thanks for fixing this

PR #1734, opened 2020-11-16
Removal of azure.historical.** i18n keys, we are using common keys

ddonahue007:
These keys are not being utilized. We are using command keys, like "units" and "unit_tooltips"

Screen Shot 2020-11-16 at 2 36 48 PM


dlabrecq (approved):
LGTM

PR #1739, opened 2020-11-18
Incorrect hIstorical usage charts

dlabrecq:
The AWS and Azure historical pages are only showing the wrong chart type. Should be showing a usage chart, not trend charts.

https://issues.redhat.com/browse/COST-746

AWS
Screen Shot 2020-11-18 at 4 03 44 PM

Azure
Screen Shot 2020-11-18 at 4 03 32 PM


ddonahue007 (approved):
LGTM

PR #1741, opened 2020-11-19
Fixed i18n key name for limits label on historical chart legend

ddonahue007:
Fixed legend for limit so it now shows date range

Screen Shot 2020-11-19 at 8 32 06 AM

Screen Shot 2020-11-19 at 8 47 09 AM


dlabrecq (approved):
Nice catch!

PR #1743, opened 2020-11-19
Rm i18n keys

ddonahue007:
Removed more i18n keys that were dead.
Also added logic to i18n test tool to automatically exclude *_plural tag if the non-plural matching tag is found


dlabrecq (approved):
LGTM

PR #1746, opened 2020-11-21
Overview chart width

dlabrecq:
Instead of splitting the screen in half, we would like to have the chart take up about 2/3 of the available width. The progress bars can be shortened to use the remaining 1/3.

Screen Shot 2020-11-20 at 9 04 16 PM


ddonahue007 (approved):
LGTM, I really like the new layout

PR #1747, opened 2020-11-21
Account for dates with empty values

dlabrecq:
Encountered a date with empty values, which may result in a disconnected line between the forecast and current month.

https://issues.redhat.com/browse/COST-736

Screen Shot 2020-11-20 at 10 14 25 PM


ddonahue007 (approved):
LGTM

PR #1751, opened 2020-11-23
OpenShift cost forecast

dlabrecq:
Created the initial Ocp cost forecast and cone of confidence.

https://issues.redhat.com/browse/COST-738

Note: There is an issue with the forecast data; thus, the forecast is currently a flat line.
See https://issues.redhat.com/browse/COST-760

Screen Shot 2020-11-23 at 12 11 19 AM


ddonahue007 (approved):
LGTM

PR #1754, opened 2020-11-24
Refactor computed items data structure

dlabrecq:
Refactored our computed items data structure to reflect the API data structure. The flat data structure we had previously is somewhat confusing, especially how we store units for each cost type. The report API data structures have evolved and it would be ideal if we updated as well.

Nothing changes visually. We should see that same overview and historical charts, tooltips, units, etc.

https://issues.redhat.com/browse/COST-765

Our computed items now mimic this API data structure
Screen Shot 2020-11-24 at 1 30 59 PM

### DIFF
--- a/src/api/forecasts/forecast.ts
+++ b/src/api/forecasts/forecast.ts
@@ -1,26 +1,26 @@
-export interface ForecastItemValue {
+export interface ForecastValue {
   units?: string;
   value?: number | string;
 }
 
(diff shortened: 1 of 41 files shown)


boaz0 (approved):

dlabrecq:
Thanks, Boaz!

ddonahue007 (commented):
LGTM

PR #1755, opened 2020-11-24
Move tags API from Report to its own type

dlabrecq:
Want to move tags out of `Report` to its own `Tag` type.

We had added tag properties to `Report` type, but it would be cleaner if the tags API used its own `Tags` type. The tags API returns a unique data structure, compared with the `Report` type.

No visual changes. The UI should continue to function normally

https://issues.redhat.com/browse/COST-766

The `Tag` type mimics this API data structure
Screen Shot 2020-11-24 at 5 17 47 PM


ddonahue007 (approved):
LGTM

boaz0 (approved):

PR #1756, opened 2020-11-24
Move org units API from Report to its own type

dlabrecq:
Want to move org units out of `Report` to its own `Org` type.

We had added org unit properties to `Report` type, but it would be cleaner if the org units API used its own `Org` type. The org units API returns a unique data structure, compared with the `Report` type.

https://issues.redhat.com/browse/COST-767

The Org type mimics this API data structure
Screen Shot 2020-11-24 at 8 59 48 PM

### DIFF
--- a/src/api/orgs/awsOrgs.test.ts
+++ b/src/api/orgs/awsOrgs.test.ts
@@ -1,0 +1,12 @@
+jest.mock('axios');
+
+import axios from 'axios';
+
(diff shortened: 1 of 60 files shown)


ddonahue007 (approved):
LGTM

boaz0 (approved):
LPTM :+1:

dlabrecq:
Thanks Boaz & Doug!

PR #1758, opened 2020-11-26
fix(costmodels): show measurement units in the select options

boaz0:
Closes COST 769

Before

image

Now

image


dlabrecq (approved):

boaz0:
Thanks @dlabrecq

PR #1760, opened 2020-12-01
percentages were showing 'NaN' for value in…

ddonahue007:
Charts were showing NaN for `% of cost` when the item value was $0.00. This is when filtering by Tag.
the cause was dividing by zero.

### **BEFORE FIX**

Screen Shot 2020-12-01 at 2 58 15 PM

Screen Shot 2020-12-01 at 3 10 03 PM

### **AFTER FIX:**

Screen Shot 2020-12-01 at 5 25 02 PM

Screen Shot 2020-12-01 at 5 23 21 PM


dlabrecq (changes requested):
Code looks good, but need to run prettier to fix the lint issues

dlabrecq (approved):

PR #1761, opened 2020-12-02
Interactive legends to store chart container via state

dlabrecq:
Refactored interactive legends to store chart container via state.

It appears if the chart container is recreated from scratch, each time the chart is redrawn, onMouseOver events are lost. If we store the container via state, then tooltips continue to be displayed properly after a data series is hidden / shown via the interactive legend.

https://issues.redhat.com/browse/COST-762

**Before**
chrome-capture

**After**
chrome-capture


ddonahue007 (approved):
LGTM

PR #1763, opened 2020-12-02
Ocp details page percentages should be based on total cost

dlabrecq:
OpenShift supplemental and infrastructure cost percentages, shown via the details table, should be based of of the total cost. Currently, the cost percentages are based on total supplemental cost and total infrastructure cost.

Also fixed a lint issue with cost models test.

https://issues.redhat.com/browse/COST-778

Screen Shot 2020-12-02 at 12 29 32 PM


ddonahue007 (approved):
LGTM

PR #1767, opened 2020-12-03
fix(costmodels): add input validation to description, name and tags

boaz0:
closes COST 781

The following verification were added to fields like `cost model name`, `cost model description`, `tag key`, `tag values` and `tag description`:
1) Making sure that the input is not exceeding a certain length.
2) ~Making sure that the input includes only valid characters~ In a second thought, this is not needed because `JSX` HTML escapes it:
> If you want to display an HTML entity within dynamic content, you will run into double escaping issues as React escapes all the strings you are displaying in order to prevent a wide range of XSS attacks by default. Read more [here](https://shripadk.github.io/react/docs/jsx-gotchas.html#html-entities)


dlabrecq (approved):

boaz0:
Did some updates:
1) change the limit to what is set in [here](https://github.com/project-koku/koku/pull/2541)
2) remove the regexp because this is not necessary. According to [React docs](https://reactjs.org/docs/jsx-in-depth.html):

> When you pass a string literal, its value is HTML-unescaped. So these two JSX expressions are equivalent:
```
<MyComponent message="&lt;3" />
<MyComponent message={'<3'} />
```

However, if we set `dangerouslySetInnerHTML` in the JSX then this is no longer valid.

dlabrecq (changes requested):
Linter failed

dlabrecq (approved):

PR #1771, opened 2020-12-08
fix(costmodels): filter rate by metric using metric labels

boaz0:
closes COST 730


dlabrecq (changes requested):
Looks good, but can you please remove the related i18n keys in en.json?

dlabrecq (approved):

PR #1772, opened 2020-12-08
Wrong tooltip label/value for cost forecast

dlabrecq:
The chart tooltip legend may show the wrong label for infrastructure data. The legend labels appear to be swapped (e.g., the value expected for "Infrastructure cost" is labeled "Cost foreccast".

This is due to reordering legend items in order to wrap items when there is not enough room. However, we can now use PatternFly's `legendAllowWrap` feature to do that for us.

**Correct tooltip labels**
Screen Shot 2020-12-08 at 12 00 45 PM

**Example of legend item responsiveness**
chrome-capture (1)

### DIFF
--- a/src/components/charts/costChart/costChart.styles.ts
+++ b/src/components/charts/costChart/costChart.styles.ts
@@ -30,7 +30,6 @@ export const chartStyles = {
     fill: chart_color_green_100.value,
     strokeWidth: 0,
   },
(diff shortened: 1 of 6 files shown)


boaz0 (approved):

ddonahue007 (approved):
LGTM

PR #1773, opened 2020-12-09
fix(costmodels): delete i18n keys for metrics/measurements

boaz0:
closes COST 756
remove unnecessary i18n keys and removing them from toolbars and tables in cost models.


dlabrecq (approved):

boaz0:
Thanks Dan!

PR #1774, opened 2020-12-09
Enable forecast feature for dev & QA only

dlabrecq:
The forecast feature should only be enabled for dev and QA users; `cost-demo` and `insights-qa`. This is a temporary change until the forecast APIs have been fully tested and ready for production.

Also set charts min y domain back to zero.


ddonahue007 (approved):
LGTM

PR #1775, opened 2020-12-10
Sync chart tooltips

dlabrecq:
This ensures chart tooltips are updated properly when a data series is hidden / shown. Although the container view is stored in state, the tooltip legend data still needs to be updated upon each render.

https://issues.redhat.com/browse/COST-806


ddonahue007 (approved):
LGTM

boaz0:
LGTM :+1:

PR #1777, opened 2020-12-14
Fetch forecast data for dev & QA only

dlabrecq:
This ensures forecast APIs are called only for the cost-demo and insights-qa users. That is, until the forecast feature is fully tested and ready for production.

https://issues.redhat.com/browse/COST-806


ddonahue007 (approved):
LGTM

PR #1780, opened 2020-12-15
"All cloud filtered by OpenShift" forecast

dlabrecq:
Added the "All cloud filtered by OpenShift" forecast

https://issues.redhat.com/browse/COST-824

Screen Shot 2020-12-14 at 10 17 04 PM


ddonahue007 (approved):
LGTM

esebesto:
Applies to all forecast graphs: there are no data for Dec-31 - forecast data should be shown (they are returned by forecast api)

PR #1783, opened 2020-12-15
OpenShift infrastructure forecast

dlabrecq:
Added the infrastructure forecast to the OpenShift overview page. For clarity, we changed the infrastructure color from green to blue.

Note that the forecast confidence is still shown in the chart, but it has been omitted from the legend items. This was done because the legend is becoming too crowded.

When you hover over the legend item, it highlights just the forecast line. However, when you click on the legend item, it hides both the forecast line and forecast confidence cone.

The tooltip still shows both forecast and confidence, but the confidence cone appears as the last item in the list. The tooltip and legend items must appear in the same order so the indexes match exactly for on-hover events.

https://issues.redhat.com/browse/COST-817

chrome-capture 2 14 38 PM

### DIFF
--- a/src/api/forecasts/forecast.ts
+++ b/src/api/forecasts/forecast.ts
@@ -43,6 +43,7 @@ export interface Forecast {
 // eslint-disable-next-line no-shadow
 export const enum ForecastType {
   cost = 'cost',
+  infrastructure = 'infrastructure',
   supplementary = 'supplementary',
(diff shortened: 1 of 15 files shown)


boaz0 (approved):
LGTM :+1:

dlabrecq:
Thanks, Boaz!

PR #1786, opened 2020-12-18
Unable to assign a source to a cost model via the UI

dlabrecq:
The problem is that the button is disabled via testing `fetchingSourcesError` for null, while the value is assigned an empty string.

https://issues.redhat.com/browse/COST-841

Screen Shot 2020-12-18 at 10 33 12 AM


adberglund (approved):

ddonahue007 (approved):
LGTM

PR #1788, opened 2020-12-29
refactor(costmodels): cost models page

boaz0:
- show search term in the browser's URL bar
 - a better hooks implementation
 - make components more functional
 - add more tests to components


package.json
```diff
@@ -77,7 +78,7 @@
     "react-dom": "16.8.6",
     "react-i18next": "11.7.3",
     "react-redux": "6.0.0",
-    "react-router-dom": "4.3.1",
+    "react-router-dom": "5.2.0",
```
dlabrecq, line 81: I suspect React router is updated for hooks. We've had navigation issues with the Insights Chrome when attempting to update this in the past. Just want to double check that all navigation is working as expected?

src/locales/en.json
```diff
@@ -886,6 +886,30 @@
...
+    "delete_cost_model": "Delete cost model",
+    "delete_title": "Delete cost model",
+    "description": "Description",
+    "filter_by_name": "Filter by name",
+    "filter_by_source_type": "Filter by source type",
+    "header_popover": "A cost model allows you to associate a price to metrics provided by your sources to charge for utilization of resources.",
+    "header_title": "Cost models",
+    "last_change": "Last change",
+    "name": "Name",
+    "no_cost_models_description": "Create a cost model to start calculating your hybrid cloud costs using custom price lists, markups, or both. Click on the button below to begin the journey.",
```
dlabrecq, line 908: It looks like a lot of these strings have been duplicated in en.json. Are both still in use?

We've been trying to clean up dead strings, so want to ensure we're not leaving unused strings elsewhere.

dlabrecq (changes requested):
Do we have an issue for this? I'd just like to better understand why we're making the changes.

1. Why do we need the search terms in the browser URL.
2. Do we need hooks?

I'm not necessarily against using hooks, but concerned about rewriting so much production code.

boaz0:
Hi @dlabrecq 

First of all, you're right, I forgot to open a ticket for this. In regards to your other questions and concerns:

1) I saw in the details pages when the user filters results the URL is changed and contains the parameters that are used to filter the data so she/he can copy it and send it to someone else to review or to keep it in her/his bookmarks. I want to do the same here.

2) Nope, I had no intention to move from Redux to hooks. I just wanted to redundant the amount of state on the Redux store. Honestly, now that I see it, I can get the cost models from Redux and also the permissions although instead of doing API call to the RBAC I will use the insight-chrome object. The dialog states, I can also use from Redux. But other states I rather have from hooks unless you want me to use only Redux?

Thanks.

package.json
boaz0: Yeah, I did remember we had that issue. Anyway, that's not a problem. I can always use "withRouter".

src/locales/en.json
boaz0: Right, I will remove old keys :+1:

dlabrecq:
The details page uses query params, mainly so the URL can be bookmarked. We thought that would be a good idea when we first started, but it wasn't a requirement. Just curious if this was a new requirement, but doesn't sound like it.

I'm basically looking for an issue to be created. Considering all the changes, QA needs an issue in order to test.

boaz0:
@dlabrecq -> opened ticket COST-860

dlabrecq (approved):

dlabrecq (approved):
Looks good to me

PR #1789, opened 2020-12-30
feat(costmodels): sort by metric/measurement in rate table

boaz0:
Sort price list by rate metric/measurement.


src/pages/costModels/components/rateTable.tsx
```diff
@@ -14,54 +25,68 @@ interface RateTableProps extends WithT {
...
+    setSortBy({ index, direction });
+  };
   let counter = 0;
-  const rows = tiers.reduce((acc, tier, ix) => {
-    const rateKind = tier.tiered_rates ? 'regular' : 'tagging';
-    let compoundRows = [];
-    if (rateKind === 'tagging') {
-      compoundRows = [
+  const rows = tiers
+    .sort((r1, r2) => {
```
dlabrecq, line 41: If this table was only used in the wizard, we could sort client-side because the user enters the data manually. However, I believe this same table is also used for the price list.

For the price list table, sorting must be performed server-side. This is especially for a paginated table; otherwise, we may not have all the data needed to sort properly sort. For example, if the user is only viewing 10 rows at a time, but there are multiple pages of data that has yet to be retrieved.

dlabrecq (changes requested):

src/pages/costModels/components/rateTable.tsx
boaz0: @dlabrecq Yes, you're right. This table is used both in the wizard and in the cost model page. However, when showing the prices list in the cost model page, the rates are not paginated and so we can do it on client side.

dlabrecq: As long as we have all the data client-side, that sounds ok

dlabrecq (approved):

boaz0:
Sweet! Thanks Dan!

PR #1791, opened 2021-01-05
Daily chart for "All OpenShift cost" overview

dlabrecq:
Created a daily cost comparison for the "All OpenShift cost" overview page. The cumulative cost chart is sown by default, with a dropdown to select the daily cost chart.

Unfortunately, the fake (nise) data were using for the cost-demo user doesn't produce a pretty chart. The high cost values scale the chart so that it's difficult to view the infrastructure cost and forecast values without hiding other data. 

Scaling issues are nothing new, but wanted to explain why the mock looks different. I expect the chart to look closer to the mock with real data from prod.

https://issues.redhat.com/browse/COST-834

chrome-capture (1)

### DIFF
--- a/src/components/charts/common/chartUtils.ts
+++ b/src/components/charts/common/chartUtils.ts
@@ -223,10 +223,6 @@ export function fillChartDatums(datums: ChartDatum[], type: ChartType = ChartTyp
   if (!datums || datums.length === 0) {
     return result;
   }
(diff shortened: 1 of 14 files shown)


boaz0 (approved):

ddonahue007 (approved):
LGTM, nice job

PR #1792, opened 2021-01-07
Daily trend chart

dlabrecq:
Created a daily trend comparison for the remaining overview pages. The cumulative cost chart is shown by default, with a dropdown to select the daily cost chart.

This particular chart is used on the following overview pages:

- OpenShift supplementary
- All Cloud filtered by OpenShift cost
- Amazon Web Services cost
- Amazon Web Services filtered by OpenShift cost
- Microsoft Azure cost
- Microsoft Azure filtered by OpenShift cost

aws bar

Note there is a known issue with the interactive legend and chart bars, where styles are not always restored after hiding legend items. I'll create a PatternFly issue and work out a solution in the upstream repo.

https://issues.redhat.com/browse/COST-852


ddonahue007 (approved):
LGTM

PR #1797, opened 2021-01-18
chore(packages): upgrade axios and jest

boaz0:


dlabrecq (approved):
Looks ok, but curious if this is blocking something?

PR #1803, opened 2021-01-26
Dynamically hide Insights navigation links

dlabrecq:
As a user, I want page links (e.g., 'OpenShift', 'Infrastructure' > 'AWS', etc.) in the Insights navigation pane to be hidden when there are no sources and/or I do not have permissions to view data.

In addition, the UI will make use of the new user-access API instead of calling its permissions util. That will ensure the UI and back-end are always in sync.

**cost-complex user**
Screen Shot 2021-01-25 at 6 58 59 PM


ddonahue007 (approved):
LGTM, this is a nice feature

boaz0 (approved):

PR #1809, opened 2021-01-27
PermissionsBase state warning

dlabrecq:
While developing locally with ci-beta, the warning below appears in the browser console. To eliminate the warning, I've moved permissions functionality to its own component

"Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
in Async (created by PermissionsBase)"

https://issues.redhat.com/browse/COST-926


ddonahue007 (approved):
LGTM

PR #1811, opened 2021-01-27
Bottom pagination controls to not be compact

dlabrecq:
Currently, the details pages for compact pagination controls at the top and bottom of each table. We don't want the controls at the bottom to be compact, only the top.

Also moved the loading icon to table body in order to help avoid some page flashing.

https://issues.redhat.com/browse/COST-913

Screen Shot 2021-01-27 at 9 05 28 AM


ddonahue007 (approved):
LGTM

PR #1813, opened 2021-01-27
Initialize perspective menu

dlabrecq:
Need to initialize the perspective menu in the overiew's `componentDidMount` function. This will ensure the menu is reset properly when navigating between pages.

https://issues.redhat.com/browse/COST-933


ddonahue007 (approved):
LGTM

PR #1817, opened 2021-01-28
No data state

dlabrecq:
This adds a no data state to help indicate that a source provider has been created, but we're still processing their data. This makes use of the new `current_month_data` flag now available from the sources API response.

https://issues.redhat.com/browse/COST-914

**Details pages**
details no data state

**Overview tabs**
overview no data state


ddonahue007 (approved):
LGTM

PR #1819, opened 2021-01-29
Move common functions to chartUtils

dlabrecq:
The charts have a few common functions; for example, `getDomain`, `getChartNames`, etc.

We have a `chartUtils` file, which is already loaded with chart datum related functions (e.g., used by pages to transform cost reports to usable data.) I renamed this file as `chartDatumUtils` and created a new `chartUtils` for functions used by the charts only.

https://issues.redhat.com/browse/COST-942


ddonahue007 (approved):
LGTM

PR #1820, opened 2021-01-29
Enable forecast for all users

dlabrecq:
The forecast is currently limited to the cost-demo, insights-qa, and cost-management users. We would like to remove this requirement, but waiting on final testing from QA.

https://issues.redhat.com/browse/COST-947


ddonahue007 (approved):
LGTM

PR #1823, opened 2021-01-29
Move common functions to detailsUtils

dlabrecq:
The details pages have a couple common functions; for example, `handleFilterAdded`, `handleFilterRemoved`, etc. I'd like to move some of this dup code to a new file, named `detailsUtils`.

https://issues.redhat.com/browse/COST-949


ddonahue007 (approved):
LGTM

PR #1831, opened 2021-02-03
Historical Cost Explorer

dlabrecq:
Created initial Historical Cost Explorer page, based on AWS for now. More updates to follow...

- The page is only accessible via a direct URL -- Insights nav will be updated after user-access API is ready.
- Refactored some existing code so the data table can show daily data. 
- Refactored existing details components to allow filter controls to be displayed in header.

Screen Shot 2021-02-03 at 12 14 54 PM


ddonahue007 (approved):
LGTM

PR #1832, opened 2021-02-03
Details pages to show IDs, along with alias

dlabrecq:
Currently, the AWS details / breakdown pages show an ID underneath the (cluster / account) name. However, OCP, Azure, and GCP do not. This update makes all the pages behave similarly, providing there is an alias in use.

https://issues.redhat.com/browse/COST-955

**OCP Details**
ocp details

**OCP breakdown**
ocp breakdown


ddonahue007 (approved):
LGTM

PR #1834, opened 2021-02-05
Add chart to explorer page

dlabrecq:
Added historical chart to explorer page.

chrome-capture (1)


ddonahue007 (approved):
need to fix linting issue, after that LGTM. Awesome job on the chart.

PR #1836, opened 2021-02-05
Update documentation links

dlabrecq:
Update documentation links for the Cost Management UI. Note that this does not include the setting page, which is a different repo

https://issues.redhat.com/browse/COST-961


ddonahue007 (approved):
LGTM

PR #1842, opened 2021-02-10
Historical Cost Explorer

dlabrecq:
Some refactoring to remove the last AWS references. The table, chart, and filter should all reflect the selected perspective now.

- Show dynamic chart titles per selected perspective
- Filter should match specific group by options for OCP, Azure, GCP, etc.
- Table should not use AWS computed report items


ddonahue007:
LGTM

ddonahue007 (approved):

PR #1846, opened 2021-02-11
I18n dupes total

ddonahue007:
added in sub totals for duplicates found

Screen Shot 2021-02-11 at 3 23 50 PM

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -1007,7 +1007,6 @@
     "gb": "GB",
     "gb-hours": "GB-hours",
     "gb-mo": "GB-month",
-    "hours": "hours",
     "hrs": "hours",
     "tag-mo": "Tag-month",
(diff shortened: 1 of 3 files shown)


dlabrecq (approved):

src/locales/en.json
```diff
@@ -1007,7 +1007,6 @@
     "gb": "GB",
     "gb-hours": "GB-hours",
     "gb-mo": "GB-month",
-    "hours": "hours",
```
dlabrecq, line 1010: We need to leave this one here. It's not dead, but a value from an API

dlabrecq (changes requested):

src/locales/en.json
ddonahue007: done

dlabrecq (approved):

PR #1847, opened 2021-02-11
Explorer: Account for missing data

dlabrecq:
The cost reports have many missing days of data. As a result, the data shown may not appear in the correct table column.

This change ensures each column has the correct data and fills in any missing data with a "no data" message.

Screen Shot 2021-02-11 at 4 51 11 PM


ddonahue007 (approved):
LGTM

PR #1848, opened 2021-02-12
915 group by

dlabrecq:
Added some fixes to the explorer's perspective and group-by dropdown menus.

- Set explorer’s default perspective option
- Reset explorer group by tag dropdown
- Reset explorer group by org unit dropdown
- Removed x-domain so bars spread out across range
- Adjust y-axis labels by adding y-domain when max values are zero

Screen Shot 2021-02-11 at 9 47 33 PM


ddonahue007 (approved):
LGTM

PR #1849, opened 2021-02-12
Historical cost explorer tick values

dlabrecq:
This change ensures the explorer chart shows every 3rd tick value per our mocks. Should provide more room when displaying a lot of data.

Fixed some other issues below

- Adjust explorer chart container height
- Truncate long resource names shown in explorer chart
- Group by tag option should remain visible while selection is valid
- Reset filter tag selection when new group by is chosen

Screen Shot 2021-02-12 at 10 42 14 AM


ddonahue007 (approved):
LGTM

PR #1850, opened 2021-02-14
Horizontal Cost Explorer

dlabrecq:
Added horizontal scroll to explorer table. This change ensures the first two columns remain sticky, while the remainder of the table may be scrolled horizontally.

chrome-capture


ddonahue007 (approved):
LGTM

PR #1853, opened 2021-02-15
Update @redhat-cloud-services packages

dlabrecq:
Updated @redhat-cloud-services packages to latest

https://issues.redhat.com/browse/COST-909


src/api/api.ts
```diff
@@ -1,4 +1,4 @@
-import { authInterceptor as insightsAuthInterceptor } from '@redhat-cloud-services/frontend-components-utilities/files/interceptors';
+import { authInterceptor as insightsAuthInterceptor } from '@redhat-cloud-services/frontend-components-utilities/esm/interceptors';
```
boaz0, line 1: It is supposed to be `@redhat-cloud-services/frontend-components-utilities/interceptors`
See here: https://github.com/RedHatInsights/frontend-components/tree/master/packages/utils#import-path-change

boaz0 (commented):
You need to update the imports according to the docs otherwise the purpose of the upgrade is missing.

src/api/api.ts
dlabrecq: Thanks, updated

ddonahue007 (approved):
LGTM

dlabrecq:
Thanks!

PR #1854, opened 2021-02-15
Azure drop down is cut off by sticky explorer table styles

dlabrecq:
Renamed the class used by the explorer table so other pages don't pick up those styles.

https://issues.redhat.com/browse/COST-1027

Screen Shot 2021-02-15 at 3 01 33 PM


ddonahue007 (approved):
LGTM

PR #1856, opened 2021-02-16
Remove babel from webpack config

dlabrecq:
Babel does not appear to be in use. Completed a full build, started the dev server, ran tests, and linter without issues.

https://issues.redhat.com/browse/COST-934


ddonahue007 (approved):
LGTM

PR #1861, opened 2021-02-17
Changed react-router imports to use react-router-dom

dlabrecq:
It looks like this change is required for the UI to work with Chrome 2.0.

https://issues.redhat.com/browse/COST-1018


ddonahue007 (commented):
LGTM

PR #1862, opened 2021-02-17
Update date-fns package to latest version

dlabrecq:
The latest date-fns package has updated a few things.

- `formatDate` was renamed `format`
- `format` uses a Date object instead of date strings
- Formatting changed from "YYYY-MM-DD" to "yyyy-MM-dd"
- `import getDate from 'date-fns/get_date'` is now `import { getDate } from 'date-fns'`


ddonahue007 (approved):
LGTM

PR #1863, opened 2021-02-17
Cost icon is not being loaded via require after updating webpack

dlabrecq:
The cost icon, used on our empty provider state page, is not being loaded via require after updating to webpack 5.x.

This change eliminates the file-loader package.
See https://webpack.js.org/guides/asset-management/#loading-images

https://issues.redhat.com/browse/COST-1049

Screen Shot 2021-02-17 at 3 59 44 PM


ddonahue007 (approved):
LGTM

PR #1864, opened 2021-02-17
Update prettier package to latest version

dlabrecq:
https://issues.redhat.com/browse/COST-1053

This adds the `arrowParens` property to match eslint. It allows functions to skip parens if they use a single parameter.

Also removed the unused `url-loader` package.


ddonahue007 (approved):
LGTM

PR #1865, opened 2021-02-18
Using resource observer to resize charts

dlabrecq:
This introduces a resource observer to resize charts. This should resize the chart when the Insights navigation pane is collapsed / expanded.

Note that we will fallback to using window resize and the Insights' NAVIGATION_TOGGLE events should this feature not be available. However, Insights has marked the NAVIGATION_TOGGLE event as deprecated, so it will likely be removed.

https://issues.redhat.com/browse/COST-1052


ddonahue007 (approved):
LGTM

PR #1866, opened 2021-02-19
Add date ranges to Historical Cost Explorer

dlabrecq:
This adds a pre-canned date ranges to the Historical Cost Explorer.

Note that the "Last 60 days" option generates an error, since we currently do not support more than 2 months of data -- Doug suggested 3 months will be available shortly.

https://issues.redhat.com/browse/COST-1062

chrome-capture


ddonahue007 (approved):
LGTM

PR #1868, opened 2021-02-21
Historical Cost Explorer fixes

dlabrecq:
Fixed a couple small issues with the Historical Cost Explorer.

- Since we're using a date range, the export should not have options for current and previous months.
- The cost explorer bulk selector dropdown is hidden by the z-index used for table scrolling.
- Check boxes are not selected when clicked.
- Update getComputedReportItems to account for AWS org units, while grouped by services and regions

https://issues.redhat.com/browse/COST-1072
https://issues.redhat.com/browse/COST-1073
https://issues.redhat.com/browse/COST-1074

**Bulk selector before**
hidden bulk selector

**Bulk selector after (with selected row)**
Screen Shot 2021-02-20 at 10 02 05 PM

**Export before**
export

**Export after**
Screen Shot 2021-02-20 at 11 36 22 PM

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -35,7 +35,8 @@
     "change_column_title": "Month over month change",
     "cost_column_title": "Cost",
     "empty_state": "Processing data to generate a list of all services that sums to a total cost...",
-    "name_column_title": "Names",
+    "name_column_title": "$t(group_by.values.{{groupBy}}) names",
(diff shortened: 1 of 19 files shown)


boaz0 (approved):

ddonahue007 (commented):
LGTM

PR #1874, opened 2021-02-23
fixed total cost label

ddonahue007:
Closes: https://issues.redhat.com/browse/COST-1083

Broken view:
Screen Shot 2021-02-23 at 9 07 31 AM

Fixed View:
Screen Shot 2021-02-23 at 9 20 34 AM


ddonahue007:
This is intentional as the "Cost" being shown here for OCP is not a "Grand Total", but a "Sub Total"

PR #1876, opened 2021-02-23
Fixed SourceMap warnings in dev mode

dlabrecq:
Not sure why we were using `eval` for dev mode and `source-map` for production -- seems backwads? We may ultimately want `eval` in production, but this will allow debugging in both cases.

Screen Shot 2021-02-22 at 8 35 14 PM


ddonahue007 (approved):
LGTM

PR #1879, opened 2021-02-23
Historical Cost Explorer fit and finish

dlabrecq:
Small fit and finish changes after reviewing with Natalie


ddonahue007 (approved):
LGTM

PR #1882, opened 2021-02-24
Historical Cost Explorer domain

dlabrecq:
Modified the explorer chart's default domain range. The previous y-axis labels appear odd when all data series are hidden via the interactive legend.

Before
before

After
Screen Shot 2021-02-23 at 10 37 16 PM


ddonahue007 (approved):
LGTM

PR #1887, opened 2021-03-01
Fix no data state for explorer

dlabrecq:
The Historical Cost Explorer is showing the no data state incorrectly. If there is at least one source provider, then we can show the explorer

Screen Shot 2021-03-01 at 9 46 13 AM


ddonahue007 (approved):
LGTM

PR #1890, opened 2021-03-02
Preserve applied filter

dlabrecq:
This preserves the applied filter when the "Group by" dropdown is changed. This includes the details pages as well as the cost explorer.

https://issues.redhat.com/browse/COST-1090

chrome-capture


ddonahue007 (approved):
LGTM

boaz0 (approved):

PR #1891, opened 2021-03-03
Removed unused OCP keys

ddonahue007:
Removed two unused i18n keys


ddonahue007:
Clsoing this PR as I have address these issues and some other in another PR

PR #1893, opened 2021-03-03
Fix chart null (no data) padding

dlabrecq:
Charts are not padding missing data correctly. It appears that some dates don't use 'T00:00:00', so the chart's null ("no data") values are off by a day.

https://issues.redhat.com/browse/COST-1106

Before
before

After
Screen Shot 2021-03-03 at 10 16 42 AM


ddonahue007 (approved):
LGTM

PR #1895, opened 2021-03-03
Omit negative chart data

dlabrecq:
The cost-demo user in production is retrieving negative forecast values. As a result, data is being shown outside the chart's y-domain.

Although the forecast API is not expected to return negative values, we can ensure data is shown only for the chart's valid domain. That is, if we encounter negative values, our ChartDatumUtils can replace that with zero instead.

https://issues.redhat.com/browse/COST-1109

**Before**
before

**After**
after


ddonahue007 (approved):
LGTM

PR #1896, opened 2021-03-04
Cost Explorer shows 31 and 61 days

dlabrecq:
~~The explorer's "Last 30 days" and "Last 60 days" features are showing 31 and 61 days. If we include today's date, then we need to subtract 29 and 59 days from today, instead.~~

After looking at more examples, I believe the current calculations are correct. I'm going to revert the change an update comments instead.

https://issues.redhat.com/browse/COST-1117


ddonahue007 (approved):
LGTM

dlabrecq:
After looking at more examples, I believe the current calculations are correct. I'm going to close this and update the issue.

PR #1897, opened 2021-03-05
Ocp widget as federated module

dlabrecq:
The OCM team would like us display an OCP overview widget on their dashboard. Instead of duplicating code in another repo, we would like to attempt sharing components via federated modules.

As a test, I created a simple OCP chart and expose that via our webpack config. Also updating to the latest frontend-components and webpack versions.

https://issues.redhat.com/browse/COST-1124

Screen Shot 2021-03-05 at 1 30 56 PM


ddonahue007 (approved):
LGTM, nice job

PR #1900, opened 2021-03-08
OCP widget as federated module

dlabrecq:
Updated federated module to handle missing i18next dependency

https://issues.redhat.com/browse/COST-1124


ddonahue007 (approved):
LGTM

PR #1902, opened 2021-03-09
Incorrect data with multiple filters

dlabrecq:
This ensures that the details and breakdown pages represent the correct cost when multiple filters are applied.

- Updated query functions to use either logical AND or logical OR
- Fixed sum calculation of costs in getComuptedReportItem
- Renamed filterBy props as groupByValue
- Added filter_by params to summary cards
- Modified breakdown querys to include group_by from previous page as a filter

https://issues.redhat.com/browse/COST-1131

**Details page with filters**
Screen Shot 2021-03-09 at 12 39 15 AM

**Breakdown page**
Screen Shot 2021-03-09 at 12 38 30 AM

### DIFF
--- a/src/api/queries/query.ts
+++ b/src/api/queries/query.ts
@@ -1,6 +1,7 @@
 import { parse, stringify } from 'qs';
 
-export const groupByPrefix = 'or:'; // logical OR ('or:') or AND ('and:') prefix for group_by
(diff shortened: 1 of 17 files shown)


ddonahue007 (approved):
LGTM

PR #1903, opened 2021-03-09
Prepare app to expose remote modules

Hyperkid123:
cc @dlabrecq 

The remote modules should be working properly with these changes. I had to re-enable the chunk mapper to allow other apps to register remote modules correctly.

And also it was necessary to update FEC dependency due to incorrect @scalprum packages versions listed in its dependencies.

An example of how to use the module can be found here: https://github.com/project-koku/koku-ui/pull/1903/files#diff-bb830d0e6ecfec0dcd1c77dd21e86fd88d6aac900a9a491d75f3e5e99892e47dR72-R77

I will remove it once I get confirmation it works for you. You can try it in a different app locally but it's not really necessary because the remote module loading is the same even if you use it in the same app that is exposing the module. Just be aware the translations might now work in apps without the proper context.

screenshot-ci foo redhat com_1337-2021 03 09-10_48_29

### DIFF
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -1,4 +1,3 @@
-import AsyncComponent from '@redhat-cloud-services/frontend-components/AsyncComponent';
 import { I18nProvider } from 'components/i18n';
 import Maintenance from 'pages/state/maintenance';
 import React from 'react';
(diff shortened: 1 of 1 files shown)


dlabrecq:
Thank you @Hyperkid123 for providing these updates. I've tested the changes locally and would like to merge this PR. I'll just remove the example in `app.tsx` first.

Screen Shot 2021-03-09 at 12 19 01 PM

PR #1907, opened 2021-03-10
No data states should check new previous_month_data and has_data flags

dlabrecq:
When displaying the "no data" state, we currently only check the `current_month_data` flag of the sources API. However, we can still show the overview page if there is data for the previous month. If there is no data for either month, then we should display an empty state.

- For the overview page, check the new `previous_month_data` flag; as well as, `current_month_data`.
- For the details pages, we only need to check the `current_month_data` flag.
- For the cost explorer, we shall check the `has_data` flag, since the date range may cover more than just the current and previous month.

### DIFF
--- a/src/api/providers.ts
+++ b/src/api/providers.ts
@@ -39,7 +39,9 @@ export interface Provider {
   cost_models?: ProviderCostModel[];
   current_month_data?: boolean;
   customer?: ProviderCustomer;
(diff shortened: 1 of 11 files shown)


ddonahue007:
LGTM

ddonahue007 (approved):

PR #1908, opened 2021-03-10
Reset export modal's default state upon close

dlabrecq:
When the modal is opened, we're setting a default state. However, if the modal is opened a second time, it appears to lose its event handlers for `resolution` and `timeScope`.

This change resets the default state only when the modal is closed.

https://issues.redhat.com/browse/COST-1134


ddonahue007 (approved):
LGTM

PR #1910, opened 2021-03-10
Replace PageHeaderTitle with frontend-components

dlabrecq:
Update the @redhat-cloud-services/frontend-components package and remove our temporary PageHeaderTitle component.

That component was added due to a typescript error with the current package.

https://issues.redhat.com/browse/COST-1146


ddonahue007 (approved):
LGTM

PR #1911, opened 2021-03-10
Apply filter_by using filter instead of group_by params

dlabrecq:
We have a table pagination issue when making cost API requests using something like `group_by[service]=Amazon&group_by[region]=us`.

This splits costs for `AmazonS3` between regions, so the table displays multiple `AmazonS3` resource names -- don't want that breakdown.

The details pages should use something more like `group_by[service]=*&filter[service]=Amazon&filter[region]=us`

https://issues.redhat.com/browse/COST-1132


ddonahue007 (approved):
LGTM

PR #1912, opened 2021-03-12
Incorrect data with multiple filters

dlabrecq:
This change to the breakdown page shall omit filters associated with the current group_by in the details page.

- Drop filers associated with the current group_by for API requests in the breakdown page
- Renamed getLogicalOrQuery and getLogicalAndQuery functions as getQuery
- Moved getBreakdownLink function to utils

https://issues.redhat.com/browse/COST-1131


ddonahue007 (commented):
LGTM

PR #1913, opened 2021-03-12
Reset ModalSubmit state upon close

dlabrecq:
This change resets state for `ModalSubmit` upon close. Also added the current resolution, along with the current and previous months, to the filename.

Screen Shot 2021-03-12 at 12 07 33 AM

https://issues.redhat.com/browse/COST-1134


ddonahue007 (commented):
LGTM

PR #1916, opened 2021-03-16
Logical AND/OR prefix

dlabrecq:
Omit default logical `or:` and `and:` prefixes.

At one time, adding a the logical `or:` prefix was required when using multiple group_by and filter params. However, this no longer seems to be a requirement. Since logical OR is applied by default, we should be able to remove the logical `or:` prefix.

Instead of something like this:

`&filter[or:project]=boston&filter[or:project]=raleigh&filter[or:cluster]=ocp
`
the params can be simplified as:

`&filter[project]=boston&filter[project]=raleigh&filter[cluster]=ocp`


ddonahue007 (approved):
LGTM

PR #1918, opened 2021-03-16
GCP and IBM should use beta flag

dlabrecq:
GCP and IBM currently appear in the overview perspective menu for the stage environment.

For the Insights nav, we use type=beta with the user-access API to restrict access to the CI and QA environments. However, we're not currently using the param in the overview page itself.

https://issues.redhat.com/browse/COST-1172


ddonahue007 (approved):
LGTM

PR #1919, opened 2021-03-16
Refactored user-access utils

dlabrecq:
Refactored user-access utils to a common area.

This will ensure all pages are testing user-access API props in the same way.
We also won't need to rewrite code for use with the `beta=true` flag.

https://issues.redhat.com/browse/COST-1172


ddonahue007 (approved):
LGTM

PR #1924, opened 2021-03-18
Cost Explorer empty state

dlabrecq:
The Cost Explorer page is showing the "no providers" empty state incorrectly. For example, if we don't have a provider for GCP or IBM. It should only show that state if there are absolutely no providers available.

https://issues.redhat.com/browse/COST-1195


ddonahue007 (approved):
LGTM

PR #1926, opened 2021-03-19
InteractiveSources should fetch IBM sources

dlabrecq:
The InteractiveSources component is not fetching IBM providers.

We use this component at the top level of the page to show alerts for inactive source providers. It should call the sources API with type=IBM, but that's missing.

https://issues.redhat.com/browse/COST-1198


ddonahue007 (approved):
LGTM

PR #1929, opened 2021-03-23
Explorer default should be based on page permissions

dlabrecq:
Modified the Cost Explorer default perspective to be based on user RBAC permissions (i.e., using our UserAccess utils)

https://issues.redhat.com/browse/COST-1205


ddonahue007 (approved):
LGTM

PR #1930, opened 2021-03-24
changes for "ResizeObserver loop limit exceeded"

ddonahue007:
Closes: https://issues.redhat.com/browse/COST-1200

These changes are to fix an intermittent failure that "Sentry" is reporting

### DIFF
--- a/src/components/charts/costChart/costChart.tsx
+++ b/src/components/charts/costChart/costChart.tsx
@@ -17,6 +17,7 @@ import {
   getChartNames,
   getDomain,
   getLegendData,
+  getResizeObserver,
   getTooltipLabel,
   initHiddenSeries,
   isDataAvailable,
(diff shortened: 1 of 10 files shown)


dlabrecq (changes requested):

src/components/charts/costChart/costChart.tsx
```diff
@@ -59,16 +60,16 @@ interface State {
...
+
   public state: State = {
     hiddenSeries: new Set(),
     width: 0,
   };
 
   public componentDidMount() {
     this.initDatum();
-    this.initResizeObserve();
+    getResizeObserver(this.containerRef.current, this.handleResize);
```
dlabrecq, line 72: Need to save the observer, so we can release it later.

`this.observer = getResizeObserver(this.containerRef.current, this.handleResize);`

dlabrecq (changes requested):

dlabrecq (approved):
LGTM. Thanks for moving all that code

PR #1931, opened 2021-03-24
Fix base name for new /beta/openshift/cost-management path

dlabrecq:
Cost Management has been moved under OpenShift. However, the Insights nav does not work as expected (e.g., with /beta/openshift/cost-management as the URL).

When we provide the router with a base name, we're not accounting for the new "openshift" path.

https://issues.redhat.com/browse/COST-1221


ddonahue007 (approved):
LGTM

PR #1934, opened 2021-03-26
Remove beta flag for GCP

dlabrecq:
This removes special code in the UI that calls the `user-access` API with `type=GCP` and `beta=true`. Since we no longer need the `beta=true` flag, we can call `user-access` with `type=` and get all providers via a single request.

Note that we still need to update the Insights nav in various CI environments.

https://issues.redhat.com/browse/COST-1225


ddonahue007 (approved):
LGTM

dlabrecq:
Merging now so we can confirm Insights nav changes are working in ci-beta

PR #1943, opened 2021-04-12
combine key/value pairs to remove duplicates

ddonahue007:
issue: https://issues.redhat.com/browse/COST-727

cleanup of duplicates in breakdown area

Screen Shot 2021-04-12 at 8 13 03 AM


ddonahue007:
We have decided to keep keys as is for future changes...

PR #1944, opened 2021-04-13
fix for cost model filter search input

ddonahue007:
closes: https://issues.redhat.com/browse/COST-1274

These changes fix the deformation of the toolbar when you select the search input to filter by Name or Description.

https://user-images.githubusercontent.com/57504257/114562159-408ec000-9c3c-11eb-9996-849dff375da3.mov

### DIFF
--- a/src/pages/costModels/costModelsDetails/utils/filters.tsx
+++ b/src/pages/costModels/costModelsDetails/utils/filters.tsx
@@ -2,12 +2,15 @@ import {
   Dropdown,
   DropdownItem,
   DropdownToggle,
-  SearchInput,
+  InputGroup,
+  InputGroupText,
+  TextInput,
   Toolbar,
   ToolbarFilter,
   ToolbarFilterProps,
   ToolbarProps,
 } from '@patternfly/react-core';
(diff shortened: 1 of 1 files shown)


boaz0 (approved):
LGTM :+1:

dlabrecq (changes requested):
We should clarify if we should be using PatternFly's `SearchInput` component here:
https://www.patternfly.org/v4/components/search-input/design-guidelines

The magnifying glass icon appears on the left side with PatternFly, while ours appear on the right. There is also a clear icon that appears on the right, which we don't have

ddonahue007:
I am closing this PR in favor of using patternfly's component

ddonahue007:
As of right now it looks like the majority of search components are using the format based on the following: https://www.patternfly.org/v4/components/toolbar

reopening this PR so that this search will follow the same format that we are using through out cost management and the majority of the platform

dlabrecq (approved):
After reading the PatternFly guidelines below
https://www.patternfly.org/v4/guidelines/filters/

I believe we should be using "Input group". Our APIs don't support type-ahead, which is the main use case for search.

PR #1946, opened 2021-04-14
Breakdown chart should use ResizeObserver

dlabrecq:
Updated the usage chart, shown in the breakdown page, to use the our `ResizeObserver` util. 

I also removed the resize listener from the cost chart, shown in the same breakdown page. It isn't necessary to resize a pie chart, since it does not span the entire width of the card.


ddonahue007 (approved):
LGTM

PR #1948, opened 2021-04-15
Items in the "Add source" dialog should remain disabled

dlabrecq:
If a source is disabled, the "Add source" dialog's "select-all" checkbox should not be able to toggle that selection. Otherwise, users encounter an error when a source has already been assigned to a different cost model.

https://issues.redhat.com/browse/COST-1290

Source assigned to same cost model
cost-model1

Source assigned to a different cost model
cost-model2


ddonahue007 (approved):
LGTM

PR #1951, opened 2021-04-20
Upgrade webpack dev server

ddonahue007:
Updated webpack-dev-server version as well as config changes. 

The changes shoudl help in debug mode and also get rid of the websocket errors when running dev server


dlabrecq (changes requested):
Need to update the manifest

dlabrecq (approved):

PR #1952, opened 2021-04-21
Added new URLs to spandx

dlabrecq:
Insights moved Cost Management under OpenShift, so the spandx needs to be updated. Note that the old URLs still work for certain CI environments.


ddonahue007 (approved):
LGTM

PR #1949, opened 2021-04-19
Explorer chart legend items should use account aliases

dlabrecq:
This ensures that the Explorer chart legend displays account aliases.

https://issues.redhat.com/browse/COST-1316

Screen Shot 2021-04-19 at 11 19 42 AM

PR #1954, opened 2021-04-21
Ensure charts save project alias

dlabrecq:
This is a follow up to https://github.com/project-koku/koku-ui/pull/1949. Want to ensure charts save the project alias, similar to the Cost Explorer chart. 

https://issues.redhat.com/browse/COST-1316


ddonahue007 (approved):
LGTM

PR #1957, opened 2021-04-23
Replaced start:dev target with start:beta

dlabrecq:
Replaced the `start:dev` target with `start:beta` to make it clear what environment devs are running locally. This mimics the `start:beta` target of the Insights starter app.

When accessing `https://ci.foo.redhat.com:1337/beta/openshift/cost-management/`; for example, we should run the `start:beta` target. However, when accessing `https://ci.foo.redhat.com:1337/openshift/cost-management/`, we must run the `start` target. Otherwise, `fed-mods.json` is not located via webpack's public path.


ddonahue007 (approved):
LGTM, This configuration/naming change is heading in the direction of what I am testing out locally for our dev env.

PR #1958, opened 2021-04-26
Omit empty strings returned by APIs

dlabrecq:
As a workaround, I've updated the UI to ignore empty account_alias strings. However, it would be ideal if the API did not return empty strings.

https://issues.redhat.com/browse/COST-1326


ddonahue007 (approved):
LGTM

PR #1960, opened 2021-04-28
added i18n exclude list key validation

ddonahue007:
Closes: https://issues.redhat.com/browse/COST-1346

Add in i18n exclude list key validation.  This tooling will alow us to make sure any key value we have in the exclude list is an actual i18n key.

Example of output:
Screen Shot 2021-04-28 at 10 24 57 AM


dlabrecq (approved):

boaz0 (approved):
LGTM :+1:

PR #1962, opened 2021-04-28
Cost Explorer perspective

dlabrecq:
Fixed default perspective menu for Cost Explorer.

https://issues.redhat.com/browse/COST-1339
https://issues.redhat.com/browse/COST-1340


ddonahue007 (approved):
LGTM

PR #1964, opened 2021-05-07
Cost Explorer to show infrastructure cost for "OpenShift usage"

dlabrecq:
The Cost Explorer should show infrastructure cost for the "OpenShift usage" perspective. Currently, we're showing the same cost as "All OpenShift".

https://issues.redhat.com/browse/COST-1371


ddonahue007 (approved):
LGTM

PR #1966, opened 2021-05-10
Add GCP and IBM to inactive sources

dlabrecq:
The inactive sources widget only shows problems with OCP, AWS, and Azure. We need to include GCP and IBM as well.

https://issues.redhat.com/browse/COST-1375

Screen Shot 2021-05-10 at 12 05 22 PM


ddonahue007 (approved):
LGTM

PR #1967, opened 2021-05-13
Cost Explorer bar width

dlabrecq:
This modifies the Cost Explorer to dynamically adjust bar widths based on the screen size. Spacing is divided equally so bars fill up the available space in the chart.

https://issues.redhat.com/browse/COST-1383

chrome-capture


ddonahue007 (approved):
LGTM

PR #1969, opened 2021-05-14
Clear sort when group_by is updated

dlabrecq:
Since each perspective (i.e., OCP, AWS, etc.) has unique order_by values (i.e., account Vs account_alias, etc.), we need to clear the Cost Explorer sort whenever a new perspective or group_by is selected. Otherwise, the order_by value may not be valid for new API requests.

https://issues.redhat.com/browse/COST-1411


ddonahue007 (approved):
LGTM

PR #1968, opened 2021-05-14
PatternFly milestone update

dlabrecq:
- Updated PatternFly to latest milestone release
- Updated Insights frontend components
- Updated dependencies to resolve Dependabot vulnerability alert

PR #1970, opened 2021-05-17
Updated css-loader package

ddonahue007:
Closes: https://issues.redhat.com/browse/COST-1403

Updated css-loader package to resolve some depend bo security concerns


dlabrecq:
I updated css-loader as part of https://github.com/project-koku/koku-ui/pull/1968.

PR #1973, opened 2021-05-24
Fix chunk loading error in local env

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-12799

This pr adds the webpack proxy to the webpack config (currently its just copied from: https://github.com/RedHatInsights/frontend-components/tree/master/packages/config and added some minor tweaks because this project is not using the common config, it will be extracted to a standalone dependency to prevent any webpack versions issues).

~~@dlabrecq can you please test if you still see the webpack chunk loading issues? Currently, it does not support local API and it's setup only for ci-beta env. It will be configured properly if it helps with the problem.~~

~~Start it with `yarn start:beta` and open `https://ci.foo.redhat.com:8002/beta/openshift/cost-management` (https is required for now)~~

### Changes
- use webpack proxy instead of insights proxy for local development
- add an interactive start script to pick required development environment
- changed build directory to dist

I have kept the spadnx and local spandx configs intact so it can be still used with insights proxy.


dlabrecq:
@Hyperkid123 I tested the PR and was able to successfully debug using Chrome – the webpack chunk loading issues were not detected. I could not debug without the PR in Chrome.

Hyperkid123:
@dlabrecq is ok if I finish this PR to switch you from insights proxy to webpack proxy? It would mean using the webpack utils package (not the config one, this one just includes all the partial plugins etc.) and adding a setup for using local API. It will also enable dev server hot-reload and you will have a bit more control over the proxy config in general.

dlabrecq:
@Hyperkid123 That would be very much appreciated. Zack Allen modified the current webpack config to help get us this far.

I assume we will be required to restart the dev server in order to develop with local environments other than ci-beta; for example, qa-beta, prod-beta, etc.?

Hyperkid123:
> I assume we will be required to restart the dev server in order to develop with local environments other than ci-beta; for example, qa-beta, prod-beta, etc.?

@dlabrecq for now yes. But I will add this requirement to the list of future features.

Hyperkid123:
@dlabrecq Ok I have added the missing configuration. I have also added an interactive script to the `yarn start` command. It will allow you to pick stable/beta for chrome, ci/qa/prod for backend, and also if you want to use local API. Feel free to replace it if you want, its here for you so you don't have to initially fiddle with the webpack config.

Also, the app should still work with insights proxy. The start command will set `process.env.USE_PROXY` to `true` which will trigger the webpack proxy. So the `container:test` and `insights:proxy` commands should still work just fine.

I would like to ask you to test the local development and let me know if anything is missing or not working.

ddonahue007:
@Hyperkid123 from a first look this looks good, i am testing it locally now.

dlabrecq:
Been testing and things are looking good. However, I'm not able to reach https://prod.foo.redhat.com:8002/beta/openshift/cost-management

```
page cost-management not found, fallback to index.html
serving locally /beta/openshift/cost-management > cost-management ---> /Users/dlabrecq/Work/hcm/repos/koku-ui/dist/index.html 
```

I found this in the Chrome console, but Firefox doesn't load either:

Screen Shot 2021-05-26 at 11 00 42 AM

Hyperkid123:
@dlabrecq the prod env should be working now as well.

dlabrecq (approved):
Testing looks good. Thank you so much for doing this our team

ddonahue007 (approved):
LGTM, tested local/remote API and everything worked as it should

PR #1974, opened 2021-05-24
Hide aggregate type from Cost Explorer's export dialog

dlabrecq:
The Cost Explorer work uses the new `start_date` and `end_date` parameters introduced to the cost APIs. However, the monthly resolution was disabled for the API.

Until monthly resolution can be supported, we want to hide the aggregate type selection from the Cost Explorer's export dialog.

https://issues.redhat.com/browse/COST-1453

Screen Shot 2021-05-24 at 12 04 39 PM


ddonahue007 (approved):
LGTM

PR #1977, opened 2021-06-02
Omit "All cloud filtered by OpenShift" option

dlabrecq:
The team requested that we temporary omit "All cloud filtered by OpenShift" option from the overview and cost explorer pages.

https://issues.redhat.com/browse/COST-1483


ddonahue007 (approved):
LGTM

PR #1980, opened 2021-06-08
Added choice to startup script for using shared PatternFly modules

dlabrecq:
For testing PatternFly pre-release packages, we cannot use shared PatternFly dependencies. We also cannot use paths to local repos as a package dependency.

This update simply adds a choice to the start up script to share (default) PatternFly dependencies or not. If true, we prune PatternFly from the package dependencies shared with Insights.


ddonahue007 (approved):
LGTM

PR #1981, opened 2021-06-09
Simplify PatternFly shared modules

dlabrecq:
This simplifies webpack using `singleton: false` for PatternFly shared modules. Also added version ranges to PatternFly packages for shared module matching.


ddonahue007 (approved):
LGTM

PR #1982, opened 2021-06-09
GCP & Azure empty states have grammatical error

dlabrecq:
For Microsoft Azure and Google Cloud Platform, both of them use "an" and should use "a"

https://issues.redhat.com/browse/COST-1493


ddonahue007 (approved):
LGTM

PR #1984, opened 2021-06-09
Updated packages to resolve dependabot alerts

dlabrecq:
Updated packages to resolve some dependabot alerts. We have more vulnerability issues with our dev dependencies, but this is what I could update for now.


ddonahue007 (approved):
LGTM

PR #1985, opened 2021-06-09
Cone of confidence legend item

dlabrecq:
Updated the charts to show the forecast cone of confidence as its own legend item. I also used that new left-pointed triangle I created for PatternFly charts

https://issues.redhat.com/browse/COST-1492

**Cumulative view**
Screen Shot 2021-06-09 at 1 37 00 PM

**Daily view**
Screen Shot 2021-06-09 at 1 47 33 PM


ddonahue007 (approved):
LGTM

PR #1988, opened 2021-06-11
fix(webpack): use upstream federated config

redallen:
Fixes need to declare patternfly separate by using https://github.com/RedHatInsights/frontend-components/blob/master/packages/config-utils/federated-modules.js

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -5,6 +5,7 @@ const log = weblog({
   name: 'wds',
 });
 const proxy = require('@redhat-cloud-services/frontend-components-config-utilities/proxy');
+const federatedPlugin = require('@redhat-cloud-services/frontend-components-config-utilities/federated-modules');
 const ChunkMapperPlugin = require('@redhat-cloud-services/frontend-components-config-utilities/chunk-mapper');
 
 const HtmlWebpackPlugin = require('html-webpack-plugin');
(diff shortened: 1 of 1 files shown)


dlabrecq (changes requested):
I tested the PR locally and can use my own react-core package; however, I have some concerns.

Would we still be sharing @patternfly/react-charts? What about @patternfly/patternfly and @patternfly/react-styles? I don't see those packages defined in @redhat-cloud-services/frontend-components-config-utilities/federated-modules.

It also looks like the federated-modules script would not be using our versions of react and react-dom due to `singleton: true`. I know we use that today, but the script kinda tacks that control away (i.e., if we had an issue with whatever singleton version ends up being used). 

Ideally, we would still control the list of packages we wish to share. Perhaps provided as an array or some sort?

redallen:
> Would we still be sharing @patternfly/react-charts?

No, but your app and other apps currently aren't agreeing to share this package with you. It'd be best to add upstream.

> What about @patternfly/patternfly and @patternfly/react-styles?

There is nothing to share in `@patternfly/patternfly` as it has no `main` entry in its package.json. `@patternfly/react-styles` should be shared since it's a dependency of `@patternfly/react-core`.

> It also looks like the federated-modules script would not be using our versions of react and react-dom due to `singleton: true`.

It will use the versions you define in `package.json`. I'm not sure what `singleton: true` will change as the current config already has it.

> Ideally, we would still define the list of packages we wish to share. Perhaps provided as an array or some sort?

You can via `shared` or `exclude`. However other apps also have to agree to sharing so it's best to add these packages upstream.

dlabrecq (approved):
Ok, thanks for the update

PR #1989, opened 2021-06-15
AWS accounts type-ahead

dlabrecq:
Added typeahead support to the AWS details page and cost explorer. 

Currently only AWS accounts are supported by the `resource-types` API. In addition, the `resource-types/aws-accounts` API does not currently support account aliases.

https://issues.redhat.com/browse/COST-1494

chrome-capture


ddonahue007 (approved):
LGTM

PR #1990, opened 2021-06-16
AWS regions and services typeahead

dlabrecq:
Enables the AWS regions and services typeahead feature.

https://issues.redhat.com/browse/COST-1565

**Regions**
chrome-capture

**Services**
chrome-capture (1)


ddonahue007 (approved):
LGTM

PR #1991, opened 2021-06-16
Breakdown link's group_by param is missing tag key

dlabrecq:
The details pages generate links with `group_by[date]` instead of `group_by[tag:tag_key]`

https://issues.redhat.com/browse/COST-1510


ddonahue007 (approved):
LGTM

PR #1992, opened 2021-06-16
Cost model empty state not displaying when no models

dlabrecq:
Refactored cost models' empty state to test user filters only. 

Instead of looking at query params; `limit` and `offset`, it would be best to test if the `name`, `description`, and `source_type` have been filtered. In addition, there's no need to test `ordering` -- the empty state should be shown, regardless.

https://issues.redhat.com/browse/COST-1518

**Empty state**
Screen Shot 2021-06-15 at 11 21 11 PM

**No match**
Screen Shot 2021-06-15 at 11 22 01 PM

**Loading**
Screen Shot 2021-06-15 at 11 22 54 PM (2)

### DIFF
--- a/src/pages/costModels/costModelsDetails/utils/table.tsx
+++ b/src/pages/costModels/costModelsDetails/utils/table.tsx
@@ -32,7 +32,7 @@ export function getRowsByStateName(stateName: string, data: any) {
         heightAuto: true,
         cells: [
           {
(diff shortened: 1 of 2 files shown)


boaz0 (approved):

ddonahue007 (approved):
LGTM

PR #1993, opened 2021-06-16
Adjust chart legend responsiveness

dlabrecq:
Small tweak to adjust chart legend responsiveness for OCP supplementary view.

chrome-capture


ddonahue007 (approved):
LGTM

PR #1994, opened 2021-06-16
initial start on react-intl and charts

ddonahue007:
Initial react-intl change over for charts...


ddonahue007:
This was to share and test the seelct format options and show its limitations. Not for merging. Closing....

PR #1998, opened 2021-06-22
no-project items shouldn't be exportable

dlabrecq:
For OCP details, the "no-project" checkbox is disabled. However, the user may choose to "select-all items" or "select all items in the page" via the bulk selector.

This change omits "no-project" from the export modal.

https://issues.redhat.com/browse/COST-1580

Screen Shot 2021-06-22 at 10 15 55 AM


ddonahue007 (approved):
LGTM

PR #1999, opened 2021-06-22
Cost models tab border is too short

dlabrecq:
The tab's bottom border should extend to the page margin, similar to the overview tabs.

https://issues.redhat.com/browse/COST-1581

chrome-capture


ddonahue007 (approved):
LGTM

PR #2000, opened 2021-06-22
Cost models filter outputs a11y error message

dlabrecq:
The PatternFly TextInput requires either an ID or aria-label to be specified.

https://issues.redhat.com/browse/COST-1585


ddonahue007 (approved):
LGTM

PR #2001, opened 2021-06-24
Update config utils.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-14981

We also have to add the `useCloud` flag to use the `cloud.redhat.com` instead `console.redhat.com` until the migration happens.

cc @dlabrecq


dlabrecq (approved):
The updated package and PR are working. Thank you for the fix.

PR #2002, opened 2021-06-25
Add stage as an option for start script

dlabrecq:
Although `stage.foo.redhat.com` isn't working at the moment, this makes it possible to test the environment. That is, along with the package update in https://github.com/project-koku/koku-ui/pull/2001.


ddonahue007 (approved):
LGTM

PR #2004, opened 2021-07-08
Fix chrome navigation to overview.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-15177

TLDR: This problem cannot be fixed in chrome because that causes issues in other applications.

### Changes
- use router pathname to extract app nav pathname
  - this is the preferred way forwards
  - once new navigation is all the way in prod we will expose a hook with the current link pathname so no event listening will be necessary

### DIFF
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -8,7 +8,9 @@ import { createMapStateToProps } from 'store/common';
 
 import { Routes, routes } from './routes';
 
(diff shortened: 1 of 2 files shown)


dlabrecq (approved):
Tested locally and the Insights nav appears to be working now. That is, I can navigate to the overview page from a sub nav.

Thanks for the fix

PR #2005, opened 2021-07-08
update @redhat-cloud-services/frontend-components-config-utilities

redallen:
Tweak webpack config to work with the latest shared version of the webpack proxy.

Ci-beta is working:
image

...and ci-stable as well!
image

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -55,6 +55,7 @@ module.exports = (_env, argv) => {
   // Moved multiple entries to index.tsx in order to help speed up webpack
   const entry = path.join(srcDir, 'index.tsx');
   const useProxy = process.env.USE_PROXY === 'true';
+  const port = useProxy ? 1337 : 8002;
 
   log.info('~~~Using variables~~~');
   log.info(`isProduction: ${isProduction}`);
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -219,51 +220,45 @@ module.exports = (_env, argv) => {
...
+      port,
+      disableHostCheck: true,
+      headers: {
+        'Access-Control-Allow-Origin': '*',
+        'Access-Control-Allow-Headers': 'Origin, X-Requested-With, Content-Type, Accept, Authorization',
+      },
+      https: useProxy,
+      ...proxy({
+        port,
+        env: `ci-${isBeta ? 'beta' : 'stable'}`,
```
dlabrecq, line 240: Thanks for the update, Zack. I'll test locally.

We'll probably want to use `process.env.CLOUDOT_ENV` here Vs `ci`, so we can run against prod.

redallen: Cool. As a reminder, this is just to control which branch of `insights-chrome-build` to checkout and serve locally on the page.

dlabrecq (approved):
I tested locally, but could only switch between ci-beta and ci-stable if we clear cookies -- despite restarting webpack. 

Created https://issues.redhat.com/browse/RHCLOUD-15212 to track that issue separately.

webpack.config.js
dlabrecq: right. It's mostly to test nav changes with the local chrome

ddonahue007 (approved):
LGTM

PR #2006, opened 2021-07-09
Update webpack to include CLOUDOT_ENV

dlabrecq:
Update webpack to include `CLOUDOT_ENV` for the proxy's `env` prop. This will ensure we can run against prod-beta, prod-stable, etc. and not just CI environments.

Follow up to https://github.com/project-koku/koku-ui/pull/2005.


ddonahue007 (commented):
LGTM

PR #2008, opened 2021-07-09
React intl changeover

ddonahue007:
initial changes for react-intl

Screen Shot 2021-07-09 at 2 45 57 PM


src/utils/formatValue.ts
```diff
@@ -9,16 +10,31 @@ export type ValueFormatter = (value: number, unit?: string, options?: FormatOpti
 export const unitLookupKey = unit => {
```
dlabrecq, line 10: Not sure why we had unit here, but it should be units -- we can update this later

src/locales/messages.ts
```diff
@@ -0,0 +1,491 @@
...
+  },
+  OCPUsageDashboardCPUTitle: {
+    id: 'OCPUsageDashboardCPUTitle',
+    description: 'OpenShift CPU usage and requests',
+    defaultMessage: 'EN OpenShift CPU usage and requests',
+  },
+  PercentTotalCost: {
+    id: 'PercentTotalCost',
+    description: '{value} {units} ({percent}%)',
+    defaultMessage: 'EN {unit}{value} ({percent}%)',
```
dlabrecq, line 462: Please use `units`

dlabrecq (commented):
Just one small nit

dlabrecq (approved):

PR #2009, opened 2021-07-10
Replace Azure's VM-hours with hours

dlabrecq:
At one time, the Azure cost report did not provide units for the Overview's "Virtual machines usage" card. The API returned undefined as the value, so the UI hard-coded "VM-hours".

https://issues.redhat.com/browse/COST-1644

Screen Shot 2021-07-09 at 11 52 46 PM


ddonahue007 (approved):
LGTM

PR #2011, opened 2021-07-12
1649 type-ahead azure

dlabrecq:
Update the Azure details page and Cost Explorer to use new type-ahead features (i.e., account, region, & services) introduced via the `resource-types` API.

chrome-capture


ddonahue007 (approved):
LGTM

PR #2012, opened 2021-07-12
OCP type-ahead support

dlabrecq:
Update the OCP details page and Cost Explorer to use new typeahead features (i.e., clusters, projects, & nodes) introduced via the `resource-types` API.

chrome-capture


ddonahue007 (approved):
LGTM

PR #2013, opened 2021-07-12
GCP type-ahead support

dlabrecq:
Update the GCP details page and Cost Explorer to use new type-ahead features (i.e., account, region, services, & project) introduced via the `resource-types` API.

chrome-capture (1)


ddonahue007 (approved):
LGTM

PR #2015, opened 2021-07-12
IBM type-ahead support

dlabrecq:
Update the IBM details page and Cost Explorer to use new type-ahead features (i.e., account, region, services, & project) introduced via the `resource-types` API.

Implemented functionality using GCP APIs, for now.

chrome-capture


ddonahue007 (approved):
LGTM

PR #2016, opened 2021-07-12
Add cluster_alias to OCP type-ahead

dlabrecq:
The `cluster_alias` property was recently added to the resources-types/openshift-clusters API. This allows users to search by cluster name Vs ID only.

https://issues.redhat.com/browse/COST-1655

chrome-capture


ddonahue007 (approved):
LGTM

PR #2018, opened 2021-07-13
Rename Cost Explorer's "OpenShift usage" perspective label

dlabrecq:
The Cost Explorer perspective option should be renamed as "OpenShift infrastructure usage cost".

Currently, the Cost Explorer shows "OpenShift usage" as a perspective option; however, this view is shown under the "Infrastructure" tab of the overview page.

https://issues.redhat.com/browse/COST-1660


ddonahue007 (approved):
LGTM

PR #2019, opened 2021-07-13
Cost explorer should show infrastructure.usage values

dlabrecq:
The Cost Explorer's "OpenShift infrastructure usage cost" perspective option appears to be using `infrastructure.total` instead of `infrastructure.usage`.

Screen Shot 2021-07-13 at 5 12 49 PM

https://issues.redhat.com/browse/COST-1661


ddonahue007 (approved):
LGTM

PR #2020, opened 2021-07-13
Cost explorer group_by key may be undefined

dlabrecq:
This ensures the Cost Explorer's group_by key is not undefined.

https://issues.redhat.com/browse/COST-1662


ddonahue007 (approved):
LGTM

PR #2021, opened 2021-07-14
Forecast should include gap in the data

dlabrecq:
Currently, we connect the cone of confidence with the last day of the current month. However, it's preferred to show the gap when data is missing.

https://issues.redhat.com/browse/COST-1468

poc


ddonahue007 (approved):
LGTM

PR #2023, opened 2021-07-15
Cost Explorer reset cached perspective

dlabrecq:
Discovered an edge case where the Cost Explorer accidentally shows a cached perspective option. This ensures the perspective menu is reset.

https://issues.redhat.com/browse/COST-1669


ddonahue007 (approved):
LGTM

PR #2024, opened 2021-07-15
Cost distribution type change

ddonahue007:
add in the distribution type card and popup to cost model details

https://user-images.githubusercontent.com/57504257/125834126-3c8194c3-28c4-4eef-8aae-d0f48a0ffc80.mov

No Distribution type when not OpenShift
Screen Shot 2021-07-15 at 3 11 30 PM


src/pages/costModels/costModel/header.tsx
```diff
@@ -167,7 +167,7 @@ const Header: React.FunctionComponent<Props> = ({
             />
             <Tab
               eventKey={1}
-              title={<TabTitleText>Markup</TabTitleText>}
+              title={<TabTitleText>Cost calculations</TabTitleText>}
```
dlabrecq, line 170: Please localize

src/pages/costModels/costModel/index.tsx
```diff
@@ -126,10 +128,24 @@ class CostModelInformation extends React.Component<Props, State> {
...
+              >
+                <div style={styles.costCalculation}>
+                  <Grid hasGutter>
+                    <GridItem lg={6} id="refMarkup">
+                      <MarkupCard current={current} />
+                    </GridItem>
+                    <GridItem lg={6} id="refDistribution">
+                      <DistributionCard current={current} />
+                    </GridItem>
+                  </Grid>
```
dlabrecq, line 145: We need to hide the distribution card when the source isn't OpenShift.

In addition, I expect you will have a different grid layout to show markup only. For example, the current markup card spans the entire page width.

src/pages/costModels/costModel/updateDistributionDialog.tsx
```diff
@@ -0,0 +1,126 @@
...
+                  onChange={this.handleDistributionChange}
+                />
+                <Radio
+                  isChecked={this.state.distribution === 'memory'}
+                  name="memoryDistribution"
+                  label="Memory"
+                  id="memoryDistribution"
+                  value="memory"
+                  onChange={this.handleDistributionChange}
+                />
```
dlabrecq, line 94: We should add aria attributes here. Suspect PatternFly is outputting warnings in the browser console without them?

src/pages/costModels/costModel/distribution.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const DistributionCardBase: React.FunctionComponent<Props> = ({
+  isWritePermission,
+  setCostModelDialog,
+  current,
+  isUpdateDialogOpen,
+  t,
+}) => {
+  const [dropdownIsOpen, setDropdownIsOpen] = React.useState(false);
+  const distributionLabel = current.distribution === 'cpu' ? 'CPU' : 'Memory';
```
dlabrecq, line 40: Please localize

dlabrecq (changes requested):

src/pages/costModels/costModel/index.tsx
ddonahue007: It is hidden, it will not show up unless it is Openshift, and yes ewhen it is not open shift the markup card goes across the entire page. I'll add a screen shot

src/pages/costModels/costModel/header.tsx
ddonahue007: Done for all titles in this file

src/pages/costModels/costModel/distribution.tsx
ddonahue007: Done

src/pages/costModels/costModel/updateDistributionDialog.tsx
ddonahue007: Added

src/pages/costModels/costModel/costCalc.styles.ts
```diff
@@ -1,12 +1,13 @@
...
 import global_FontSize_xl from '@patternfly/react-tokens/dist/js/global_FontSize_xl';
-import global_spacer_lg from '@patternfly/react-tokens/dist/js/global_spacer_lg';
 import React from 'react';
 
 export const styles = {
   card: {
-    minHeight: '130px',
-    marginLeft: global_spacer_lg.value,
-    marginRight: global_spacer_lg.value,
+    minHeight: '150px',
```
dlabrecq, line 7: This `minHeight` value doesn't work with a string value. When defining styles like this, the value should just be a number `150`.

dlabrecq (changes requested):

src/pages/costModels/costModel/costCalc.styles.ts
ddonahue007: done

dlabrecq (approved):

PR #2025, opened 2021-07-16
Forecast API may result in data gap

dlabrecq:
This fills in a one-day gap between forecast and Azure's last reported day for the current month.

https://issues.redhat.com/browse/COST-1666

Screen Shot 2021-07-16 at 1 53 57 PM


ddonahue007 (approved):
LGTM

PR #2026, opened 2021-07-16
"AWS filtered by OpenShift" type-ahead support

dlabrecq:
Updated Cost Explorer's "AWS filtered by OpenShift" perspective to use new type-ahead features (i.e., account, region, services, & project) introduced via the `resource-types` API.

https://issues.redhat.com/browse/COST-1675

chrome-capture


ddonahue007 (approved):
LGTM

PR #2031, opened 2021-07-19
added markup or discount model changes

ddonahue007:
Changes for pop up modal for markup or discount changes

Screen Shot 2021-07-19 at 6 01 28 PM

### DIFF
--- a/src/pages/costModels/costModel/costCalc.styles.ts
+++ b/src/pages/costModels/costModel/costCalc.styles.ts
@@ -13,4 +13,7 @@ export const styles = {
     fontSize: global_FontSize_xl.value,
     textAlign: 'center',
   },
+  radioAlign: {
+    marginTop: 15,
+  },
 } as { [className: string]: React.CSSProperties };
--- a/src/pages/costModels/costModel/costCalc.styles.ts
+++ b/src/pages/costModels/costModel/costCalc.styles.ts
@@ -16,4 +16,10 @@ export const styles = {
   radioAlign: {
     marginTop: 15,
   },
+  sign: {
+    borderRight: 0,
+  },
+  percent: {
+    borderLeft: 0,
+  },
 } as { [className: string]: React.CSSProperties };
--- a/src/pages/costModels/costModel/updateDistributionDialog.tsx
+++ b/src/pages/costModels/costModel/updateDistributionDialog.tsx
@@ -1,10 +1,23 @@
-import { Alert, Button, Form, FormGroup, Modal, Radio, Stack, StackItem, Title } from '@patternfly/react-core';
+import {
+  Alert,
+  Button,
+  Form,
+  FormGroup,
+  Modal,
+  Radio,
+  Stack,
+  StackItem,
+  Text,
+  TextContent,
+} from '@patternfly/react-core';
 import { CostModel } from 'api/costModels';
 import React from 'react';
 import { WithTranslation, withTranslation } from 'react-i18next';
 import { connect } from 'react-redux';
 import { createMapStateToProps } from 'store/common';
 import { costModelsActions, costModelsSelectors } from 'store/costModels';
+
+import { styles } from './costCalc.styles';
 
 interface Props extends WithTranslation {
   isLoading: boolean;
@@ -69,9 +82,9 @@ class UpdateDistributionModelBase extends React.Component<Props, State> {
         <Stack hasGutter>
           <StackItem>{error && <Alert variant="danger" title={`${error}`} />}</StackItem>
           <StackItem>
-            <Title headingLevel="h2" size="md">
-              {t('cost_models_details.description_distribution_model')}
-            </Title>
+            <TextContent>
+              <Text style={styles.cardDescription}>{t('cost_models_details.description_distribution_model')}</Text>
+            </TextContent>
           </StackItem>
           <StackItem>
             <Form>
--- a/src/pages/costModels/costModel/updateDistributionDialog.tsx
+++ b/src/pages/costModels/costModel/updateDistributionDialog.tsx
@@ -91,7 +91,7 @@ class UpdateDistributionModelBase extends React.Component<Props, State> {
               <FormGroup isInline fieldId="cost-distribution" isRequired>
                 <Radio
                   isChecked={this.state.distribution === 'cpu'}
-                  name="cpuDistribution"
+                  name="distribution"
                   label={t('cpu_title')}
                   aria-label={t('cpu_title')}
                   id="cpuDistribution"
@@ -100,7 +100,7 @@ class UpdateDistributionModelBase extends React.Component<Props, State> {
                 />
                 <Radio
                   isChecked={this.state.distribution === 'memory'}
-                  name="memoryDistribution"
+                  name="distribution"
                   label={t('memory_title')}
                   aria-label={t('memory_title')}
                   id="memoryDistribution"
--- a/src/pages/costModels/costModel/updateMarkupDialog.tsx
+++ b/src/pages/costModels/costModel/updateMarkupDialog.tsx
@@ -3,15 +3,19 @@ import {
   Button,
   Form,
   FormGroup,
+  Grid,
+  GridItem,
   InputGroup,
   InputGroupText,
+  List,
+  ListItem,
   Modal,
+  Radio,
   Stack,
   StackItem,
   Text,
   TextContent,
   TextInput,
-  TextVariants,
   Title,
 } from '@patternfly/react-core';
 import { CostModel } from 'api/costModels';
@@ -21,6 +25,8 @@ import { connect } from 'react-redux';
 import { createMapStateToProps } from 'store/common';
 import { costModelsActions, costModelsSelectors } from 'store/costModels';
 import { formatValue } from 'utils/formatValue';
+
+import { styles } from './costCalc.styles';
 
 interface Props extends WithTranslation {
   isLoading: boolean;
@@ -32,23 +38,46 @@ interface Props extends WithTranslation {
 
 interface State {
   markup: string;
+  origSign: string;
+  sign: string;
 }
 
 class UpdateMarkupModelBase extends React.Component<Props, State> {
   constructor(props) {
     super(props);
+    const initialMarkup = this.props.current.markup.value;
+    const initialSign = initialMarkup.startsWith('-') ? '-' : '+';
+    const noSignValue = initialMarkup.startsWith('-') ? initialMarkup.substring(1) : initialMarkup;
+
     this.state = {
       markup:
-        (formatValue(Number(this.props.current.markup.value), 'markup', {
+        (formatValue(Number(noSignValue), 'markup', {
           fractionDigits: 2,
         }) as string) || '0.00',
+      origSign: initialSign,
+      sign: initialSign,
     };
   }
+
+  private handleSignChange = (_, event) => {
+    const { value } = event.currentTarget;
+    this.setState({ sign: value });
+  };
+
+  private handleMarkupDiscountChange = (_, event) => {
+    const { value } = event.currentTarget;
+    this.setState({ markup: value });
+  };
+
+  private markupValidator = () => {
+    return /^\d*(\.\d{1,2})?$/.test(this.state.markup) ? 'default' : 'error';
+  };
+
   public render() {
     const { error, current, onClose, updateCostModel, isLoading, t } = this.props;
     return (
       <Modal
-        title={t('cost_models_details.edit_markup')}
+        title={t('cost_models_details.edit_markup_or_discount')}
         isOpen
         onClose={() => onClose({ name: 'updateMarkup', isOpen: false })}
         variant="small"
@@ -62,7 +91,7 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
                 source_uuids: current.sources.map(provider => provider.uuid),
                 source_type: current.source_type === 'OpenShift Container Platform' ? 'OCP' : 'AWS',
                 markup: {
-                  value: this.state.markup,
+                  value: this.state.sign === '-' ? this.state.sign + this.state.markup : this.state.markup,
                   unit: 'percent',
                 },
               };
@@ -70,11 +99,15 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
             }}
             isDisabled={
               isNaN(Number(this.state.markup)) ||
-              Number(this.state.markup) === Number(current.markup.value || 0) ||
+              (Number(this.state.markup) ===
+                Number(
+                  current.markup.value.startsWith('-') ? current.markup.value.substring(1) : current.markup.value || 0
+                ) &&
+                this.state.sign === this.state.origSign) ||
               isLoading
             }
           >
-            {t('cost_models_details.add_rate_modal.save')}
+            {t('save')}
           </Button>,
           <Button
             key="cancel"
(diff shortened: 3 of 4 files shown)


src/pages/costModels/costModel/updateMarkupDialog.tsx
```diff
@@ -62,76 +91,111 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
...
+                    <div style={styles.radioAlign}>
+                      <Radio
+                        isChecked={this.state.sign === '-'}
+                        name="discount"
+                        label={t('cost_models_details.discount_minus')}
+                        aria-label={t('cost_models_details.discount_minus')}
+                        id="discount"
+                        value="-"
+                        onChange={this.handleSignChange}
+                      />
```
dlabrecq, line 156: Instead of maintaining a '-' and "+" value, it would be ideal to just store a boolean variable. For example, store `discount` simply as `true` or `false`.

src/pages/costModels/costModel/updateMarkupDialog.tsx
```diff
@@ -62,76 +91,111 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
...
               updateCostModel(current.uuid, newState, 'updateMarkup');
             }}
             isDisabled={
               isNaN(Number(this.state.markup)) ||
-              Number(this.state.markup) === Number(current.markup.value || 0) ||
+              (Number(this.state.markup) ===
+                Number(
+                  current.markup.value.startsWith('-') ? current.markup.value.substring(1) : current.markup.value || 0
+                ) &&
+                this.state.sign === this.state.origSign) ||
```
dlabrecq, line 106: It appears we're parsing "+" or "-" strings in a lot of places. We could test if markup is negative, then apply the sign.

src/pages/costModels/costModel/updateMarkupDialog.tsx
```diff
@@ -62,76 +91,111 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
...
-                    }}
-                    validated={!isNaN(Number(this.state.markup)) ? 'default' : 'error'}
-                  />
-                  <InputGroupText style={{ borderLeft: '0' }}>%</InputGroupText>
-                </InputGroup>
-              </FormGroup>
+              <Grid hasGutter>
+                <GridItem lg={8} id="refSign">
+                  <FormGroup isInline fieldId="markup-or-discount" label={t('cost_models_details.markup_or_discount')}>
+                    <div style={styles.radioAlign}>
```
dlabrecq, line 136: Don't use a margin to push the radio button down, align it vertically; using center or bottom.

Note that we don't need to place each radio button in an extra `div` to apply a style. `GridItem`, `FormGroup`, and `radio` accept a class name.

src/pages/costModels/costModel/updateMarkupDialog.tsx
```diff
@@ -62,76 +91,111 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
...
+                      <TextInput
+                        style={{ borderLeft: '0' }}
+                        type="text"
+                        aria-label={t('rate')}
+                        id="markup-input-box"
+                        value={this.state.markup}
+                        onChange={this.handleMarkupDiscountChange}
+                        validated={this.markupValidator()}
+                      />
+                      <InputGroupText style={{ borderLeft: '0' }}>%</InputGroupText>
```
dlabrecq, line 178: It would be best to create a class for `borderLeft: '0'` and `borderRight: '0'`. Try to avoid inline styles

src/pages/costModels/costModel/updateMarkupDialog.tsx
```diff
@@ -32,23 +38,46 @@ interface Props extends WithTranslation {
...
+  origSign: string;
+  sign: string;
 }
 
 class UpdateMarkupModelBase extends React.Component<Props, State> {
   constructor(props) {
     super(props);
+    const initialMarkup = this.props.current.markup.value;
+    const initialSign = initialMarkup.startsWith('-') ? '-' : '+';
+    const noSignValue = initialMarkup.startsWith('-') ? initialMarkup.substring(1) : initialMarkup;
```
dlabrecq, line 50: It appears we're parsing for "+" or "-" strings in multiple places. Instead of parsing strings, it would be cleaner to test if the markup is negative or positive. 

Perhaps this could be defined once, for example: `hasDiscount = Number(markup) < 0;`

Please see my radio button comments below

src/pages/costModels/costModel/updateMarkupDialog.tsx
```diff
@@ -62,76 +91,111 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
...
-            <Title headingLevel="h2" size="md">
-              {t('cost_models_details.table.columns.name')}
-            </Title>
-          </StackItem>
           <StackItem>
             <TextContent>
-              <Text component={TextVariants.h6}>{current.name}</Text>
+              <Text style={styles.cardDescription}>
+                {t('cost_models_details.description_markup_or_discount_model')}
+              </Text>
```
dlabrecq, line 128: This should probably be a `Title` because it defines a new section, like `Examples`

src/pages/costModels/costModel/updateMarkupDialog.tsx
```diff
@@ -62,76 +91,111 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
...
+              (Number(this.state.markup) ===
+                Number(
+                  current.markup.value.startsWith('-') ? current.markup.value.substring(1) : current.markup.value || 0
+                ) &&
+                this.state.sign === this.state.origSign) ||
               isLoading
             }
           >
-            {t('cost_models_details.add_rate_modal.save')}
+            {t('save')}
```
dlabrecq, line 110: Mocks say "Save changes"

dlabrecq (changes requested):

src/pages/costModels/costModel/costCalc.styles.ts
```diff
@@ -13,4 +13,7 @@ export const styles = {
     fontSize: global_FontSize_xl.value,
     textAlign: 'center',
   },
+  radioAlign: {
+    marginTop: 15,
+  },
```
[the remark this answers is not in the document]
dlabrecq: Actually, since this `div` is already in a flex box, you could use `align-self: flex-end`

src/pages/costModels/costModel/updateMarkupDialog.tsx
ddonahue007: true I can do that and I started out that way, one of the issues there was then I have to constantly convert the sign to a boolean.

src/pages/costModels/costModel/updateMarkupDialog.tsx
ddonahue007: Yes It does and I talked to Natalie about that, we decided Save was fine

src/pages/costModels/costModel/updateMarkupDialog.tsx
ddonahue007: IN some cases we are parsing what state.sign is set to and then based on that we are either adding a sign to the markup or not.  Tese test are needed now as we no longer except the sign to be adding into the text area.

src/pages/costModels/costModel/updateMarkupDialog.tsx
ddonahue007: Agreed, just missed that.

src/pages/costModels/costModel/updateMarkupDialog.tsx
dlabrecq: Since we're using the radio buttons, the text input shouldn't accept a negative value

Screen Shot 2021-07-20 at 12 17 17 PM

dlabrecq: For example, the text input can ignore certain characters.

https://stackoverflow.com/questions/14806200/disable-some-characters-from-input-field

src/pages/costModels/costModel/updateDistributionDialog.tsx
```diff
@@ -69,9 +82,9 @@ class UpdateDistributionModelBase extends React.Component<Props, State> {
         <Stack hasGutter>
           <StackItem>{error && <Alert variant="danger" title={`${error}`} />}</StackItem>
           <StackItem>
-            <Title headingLevel="h2" size="md">
-              {t('cost_models_details.description_distribution_model')}
-            </Title>
+            <TextContent>
+              <Text style={styles.cardDescription}>{t('cost_models_details.description_distribution_model')}</Text>
+            </TextContent>
```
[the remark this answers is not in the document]
dlabrecq: Ok, didn't realize this was just a description.

src/pages/costModels/costModel/updateMarkupDialog.tsx
```diff
@@ -62,76 +91,111 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
...
+                    <InputGroup style={{ width: '150px' }}>
+                      <InputGroupText style={{ borderRight: '0' }}>{this.state.sign}</InputGroupText>
+                      <TextInput
+                        style={{ borderLeft: '0' }}
+                        type="text"
+                        aria-label={t('rate')}
+                        id="markup-input-box"
+                        value={this.state.markup}
+                        onChange={this.handleMarkupDiscountChange}
+                        validated={this.markupValidator()}
```
dlabrecq, line 176: This shouldn't have parens, just `validated={this.markupValidator}`

src/pages/costModels/costModel/updateMarkupDialog.tsx
dlabrecq: For example, the text input can ignore certain characters.

```
onKeyUp={this.lettersOnly}
...
private lettersOnly = event => {
    const regex = /^[0-9]*$/;
    event.target.value = event.target.value.replace(regex, '');
  }
```

src/pages/costModels/costModel/updateMarkupDialog.tsx
ddonahue007: I guess I am a bit confused on this comment. This is the description of the modal/card. The title is above the description on line 81.

`title={t('cost_models_details.edit_markup_or_discount')}`

src/pages/costModels/costModel/updateMarkupDialog.tsx
ddonahue007: Done

dlabrecq (commented):
FYI, pressing enter in the input field submits the page without setting the rate. Probably want to catch the enter key, and call save?

src/pages/costModels/costModel/updateMarkupDialog.tsx
dlabrecq: Sorry, I thought this was the "Markup or Discount" title.

dlabrecq: It confused me because there is another description edit in an unrelated card.

ddonahue007:
> FYI, pressing enter in the input field submits the page without setting the rate. Probably want to catch the enter key, and call save?

Fixed

src/pages/costModels/costModel/updateMarkupDialog.tsx
ddonahue007: Nice trick

src/pages/costModels/costModel/updateMarkupDialog.tsx
dlabrecq: Maybe I'm misunderstanding, but you don't need to convert the boolean to a string. Unless the user clicks save.

Seems that conversion already happens via the onClick? I was thinking it would look more like:

```
markup: {
  value: this.state.discount ? '-' + this.state.markup : this.state.markup,
  unit: 'percent',
},
```

ddonahue007: Right I can added another state value, and then use both state.discount and state.sign. I thought you were asking for the sign to be replaced.

dlabrecq: The other benefit of using boolean is that you don't need to test

isChecked={this.state.sign === '-'} Vs isChecked={this.state.sign === '+'}

If you give the radiobutton the same name (i.e., discount), the radio button would be mutually exclusive. Currently, both buttons could be selected, since they use different names.

ddonahue007: its here as well
<InputGroupText style={styles.sign}>{this.state.sign}</InputGroupText>

dlabrecq: I believe this would be preferable to parsing strings in multiple places.

It's also best practice to use a boolean with the radio buttons, ensuring they are mutually exclusive.

src/pages/costModels/costModel/updateDistributionDialog.tsx
```diff
@@ -1,11 +1,24 @@
-import { Alert, Button, Form, FormGroup, Modal, Radio, Stack, StackItem, Title } from '@patternfly/react-core';
+import {
+  Alert,
+  Button,
+  Form,
+  FormGroup,
+  Modal,
+  Radio,
```
dlabrecq, line 7: Missed the radio buttons in this dialog, but they should also be mutually exclusive (i.e., use the same name)

src/pages/costModels/costModel/updateMarkupDialog.tsx
ddonahue007: OK this seems to be working

src/pages/costModels/costModel/updateDistributionDialog.tsx
ddonahue007: Done

src/pages/costModels/costModel/updateMarkupDialog.tsx
```diff
@@ -32,23 +38,52 @@ interface Props extends WithTranslation {
...
+    this.setState({ markup: value });
+  };
+
+  private markupValidator = () => {
+    return /^\d*(\.?\d{1,2})?$/.test(this.state.markup) ? 'default' : 'error';
+  };
+
+  private lettersOnly = event => {
+    const regex = /[^0-9.]/g;
+    event.target.value = event.target.value.replace(regex, '');
```
dlabrecq, line 78: We should probably use `currentTarget` Vs `target`.

dlabrecq (changes requested):
Looking good. Just a couple nits and an issue with the validator.

When I type '-', the `markupValidator` is called before `lettersOnly`. This results in the red error indicator being shown despite the '-' char being removed.

The debugger shows that `handleMarkupDiscountChange` is first to be invoked, which calls `setState`. Calling `setState` always invokes `render`, which in-turn is responsible for calling `markupValidator`.

As a quick fix, `handleMarkupDiscountChange` could test a similar regex as `lettersOnly`, before calling `setState`. For example:

```
  private handleMarkupDiscountChange = (_, event) => {
    const { value } = event.currentTarget;
    const regex = /^[0-9.]*$/;
    if (regex.test(value)) {
      this.setState({ markup: value });
    }
  };
```

src/pages/costModels/costModel/updateMarkupDialog.tsx
```diff
@@ -32,23 +38,52 @@ interface Props extends WithTranslation {
...
+  origIsDiscount: boolean;
+  isDiscount: boolean;
 }
 
 class UpdateMarkupModelBase extends React.Component<Props, State> {
   constructor(props) {
     super(props);
+    const initialMarkup = this.props.current.markup.value;
+    const isNegative = Number(initialMarkup) < 0;
+    const noSignValue = isNegative ? initialMarkup.substring(1) : initialMarkup;
```
dlabrecq, line 50: Should this have a default of zero? I ask because there is a zero default below on line 10.

    `const noSignValue = isNegative ? initialMarkup.substring(1) : initialMarkup || 0;`

src/pages/costModels/costModel/updateMarkupDialog.tsx
dlabrecq: We should consider renaming this method, since we don't accept letters (e.g., handleOnKeyDown, numbersOnly, etc.)?

dlabrecq (approved):

dlabrecq (approved):

PR #2032, opened 2021-07-20
upgrade css-loader

ddonahue007:
This upgrade will upgrade one of the version of post-css in our manifest


dlabrecq (changes requested):
Please try updating @redhat-cloud-services/frontend-components as well. See if we can get rid of the security issue.

Edit: never mind, the Insights pkg still uses postcss@8.2.7.

dlabrecq (approved):

PR #2036, opened 2021-07-21
GCP filtered by OpenShift

dlabrecq:
This stubs-out the "GCP filtered by OpenShift" overview page. Until new APIs are ready, we're using the existing GCP APIs.

https://issues.redhat.com/browse/COST-1693

Screen Shot 2021-07-21 at 2 31 21 PM


ddonahue007 (approved):
LGTM

PR #2038, opened 2021-07-21
Rename awsCloud and azureCloud components

dlabrecq:
The overview perspective contains options for "AWS filtered by OpenShift" and "Azure filtered by OpenShift". However, the components associated with these perspectives are prefixed with `awsCloud` and `azureCloud`. 

This naming convention was used when these perspectives were located in a "cloud" tab, which is no longer used. Therefore, components such as `awsCloudDashboard` and `azureCloudDashboard`; for example, should be renamed as `awsOcpDashboard` and `azureOcpDashboard`.

https://issues.redhat.com/browse/COST-1700


ddonahue007 (approved):
LGTM

PR #2040, opened 2021-07-22
API interfaces should extend PagedResponse

dlabrecq:
The `Report` interface currently defines its own types for `ReportData`, `ReportLinks`, and `ReportMeta`. However, `api.ts` also defines similar interfaces for `PagedLinks`, `PagedMetaData`, and `PagedResponse`.

With a slight modification, the `Report` interface could extend `PagedResponse`. Thus, eliminating the need for the `ReportLinks` interface.

There are similar interfaces for tags, resources, forecasts, etc.

https://issues.redhat.com/browse/COST-1708

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -12,8 +12,8 @@ export interface PagedLinks {
   last: string;
 }
 
-export interface PagedResponse<D = any> {
-  meta: PagedMetaData;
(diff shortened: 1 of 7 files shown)


ddonahue007 (approved):
LGTM

ddonahue007 (commented):
LGTM

PR #2043, opened 2021-07-26
Clean directory before releasing another branch

dlabrecq:
Need to clean up newly created .git directory before releasing a second branch


ddonahue007 (commented):
LGTM

PR #2046, opened 2021-07-27
webpack config should use useCloud=false

dlabrecq:
We want to change webpackk.config.js to `useCloud: false`, since `cloud.stage.readhat.com` is going away.

Also changed the startup script to default to using the `stage` env.

https://issues.redhat.com/browse/COST-1726


ddonahue007 (approved):

ddonahue007 (approved):
LGTM

PR #2054, opened 2021-07-29
Cost dist wizard

ddonahue007:
Closes: https://issues.redhat.com/browse/COST-1505

Added changes for mark and discount along with Distribution type(Openshift only) to the create cost model wizard.

**Cost calculations  step** no Distribution type (non-openshift)

Screen Shot 2021-07-28 at 8 37 20 PM

**review step** no Distribution type (non-openshift)

Screen Shot 2021-07-28 at 8 37 41 PM

**Cost calculation step** with Distribution type

Screen Shot 2021-07-28 at 8 35 30 PM

**review step**  with Distribution type

Screen Shot 2021-07-28 at 8 35 58 PM


src/pages/costModels/createCostModelWizard/index.tsx
```diff
@@ -333,7 +349,7 @@ class CostModelWizardBase extends React.Component<Props, State> {
             name: this.state.name,
             type: this.state.type,
             description: this.state.description,
-            markup: this.state.markup,
+            markup: this.state.isDiscount ? '-' + this.state.markup : this.state.markup,
```
dlabrecq, line 352: I'm a little confused that you're setting `markup` for the context, but not `isDiscount`. On line 74, I see that you deconstruct `isDiscount` like so:

Screen Shot 2021-07-29 at 10 28 28 AM

Why is the context different in each case?

src/pages/costModels/createCostModelWizard/markup.tsx
```diff
@@ -1,79 +1,115 @@
...
+                  {t('cost_calculations')}
                 </Title>
               </StackItem>
               <StackItem>
+                <Title headingLevel="h3" size="md">
+                  {t('cost_models_details.markup_or_discount')}
+                </Title>
                 <TextContent>
-                  <Text component={TextVariants.h6}>{t('cost_models_wizard.markup.sub_title')}</Text>
+                  <Text>{t('cost_models_wizard.description_markup_or_discount_model')}</Text>
```
dlabrecq, line 43: The reason we use `Text` is to define the `h6` tag for screen readers. We shouldn't eliminate the variant so the text isn't bold. It's still represents a new section that screen readers need to announce to users.

I know it probably matches the mocks, but it's necessary for a11y.

src/pages/costModels/createCostModelWizard/markup.tsx
```diff
@@ -1,79 +1,115 @@
...
+                    <GridItem lg={4} id="refMarkup">
+                      <FormGroup
+                        isInline
+                        fieldId="rate"
+                        helperTextInvalid={t('cost_models_wizard.markup.invalid_markup_text')}
+                      >
+                        <InputGroup style={{ width: '150px' }}>
+                          <InputGroupText style={styles.sign}>{isDiscount ? '-' : '+'}</InputGroupText>
+                          <TextInput
+                            style={{ borderLeft: '0' }}
```
dlabrecq, line 84: It would be best not to introduce inline styles -- add them to `costCalc.styles.ts`. Same goes for `updateMarkupDialog.tsx` -- mentioned that in a previous review.

src/pages/costModels/createCostModelWizard/reviewWithDistribution.tsx
```diff
@@ -0,0 +1,136 @@
...
+                      <WarningIcon text={t('cost_models_wizard.warning_override_sources')} />
+                    )}
+                  </TextListItem>
+                  <TextListItem component={TextListItemVariants.dd}>
+                    {sources
+                      .filter(r => r.selected)
+                      .map(r => r.name)
+                      .join(', ')}
+                  </TextListItem>
+                </TextList>
```
dlabrecq, line 111: This looks like an exact copy of `review.tsx` with the exception of these new lines. It would be best to conditionally include this in the existing file.

dlabrecq (changes requested):
Ran through the wizard steps and it's looking good. 

Something I missed in the previous reviews is the size of the markup dialog. We should change this from a `small` dialog to use the `medium` size variant. Then the text won't wrap and it will look a little nicer.

Screen Shot 2021-07-29 at 10 17 23 AM

src/pages/costModels/createCostModelWizard/reviewWithDistribution.tsx
ddonahue007: I agree that there is duplication of code, I was following the the structure that was already present using the source_type along with steps to decide on which component to load. I can certain rewrite the code to be conditional but seems like if we go that route that we might want to revisit how this wizard is put together and maybe do a larger refactor in another PR?

src/pages/costModels/createCostModelWizard/index.tsx
ddonahue007: Not setting isDiscount, but using the value(line 81) that has already been set in context.

src/pages/costModels/createCostModelWizard/reviewWithDistribution.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+const ReviewDetailsBase: React.SFC<WithTranslation> = ({ t }) => (
+  <CostModelContext.Consumer>
+    {({ name, description, distribution, type, markup, sources, tiers, createError, isDiscount }) => {
+      return (
+        <>
+          {createError && <Alert variant="danger" title={`${createError}`} />}
+          <Stack hasGutter>
+            <StackItem>
+              <Title headingLevel="h2" size="xl">
```
dlabrecq, line 60: Wrap with `<TextContent>`?

ddonahue007:
> Ran through the wizard steps and it's looking good.
> 
> Something I missed in the previous reviews is the size of the markup dialog. We should change this from a `small` dialog to use the `medium` size variant. Then the text won't wrap and it will look a little nicer.
> 
> Screen Shot 2021-07-29 at 10 17 23 AM

I can certainly do that, but it will no longer match the

src/pages/costModels/createCostModelWizard/markup.tsx
```diff
@@ -1,79 +1,115 @@
...
       <CostModelContext.Consumer>
-        {({ onMarkupChange, markup }) => {
+        {({ handleSignChange, handleOnKeyDown, handleMarkupDiscountChange, markupValidator, markup, isDiscount }) => {
           return (
             <Stack hasGutter>
               <StackItem>
                 <Title headingLevel="h2" size="xl">
-                  {t('cost_models_wizard.markup.title')}
+                  {t('cost_calculations')}
                 </Title>
```
dlabrecq, line 36: I believe all `<Title>` components are expected to be wrapped by `<TextContent>` to have the proper spacing. They're used in a lot of places, so I'll let you search for them.

http://patternfly-react.surge.sh/components/text/

dlabrecq (changes requested):

src/pages/costModels/createCostModelWizard/reviewWithDistribution.tsx
ddonahue007: I am using the same mechanism for these pages as well. I agree that there is dupe code here. The way it looks like it was design is to use the sources type as a switch and then based on that you have a series of steps. Those steps are components.  If you look at steps.tsx that is how pricelist and these new changes are handled.

src/pages/costModels/createCostModelWizard/markup.tsx
ddonahue007: Done

src/pages/costModels/createCostModelWizard/markup.tsx
ddonahue007: Doing a search in our code base and I do not see any <Title> tags wrapped in  <TextContent>. If we think this is a pattern we should be following i can open another issues to track a sweep of the code.

src/pages/costModels/createCostModelWizard/markup.tsx
ddonahue007: Done

src/pages/costModels/createCostModelWizard/markup.tsx
dlabrecq: Sorry, I was speaking about the `Text` component. You replaced `Title` with `Text` and got rid of the `h6`

src/pages/costModels/createCostModelWizard/markup.tsx
```diff
@@ -14,66 +19,100 @@ import { Form } from 'components/forms/form';
...
               </StackItem>
               <StackItem>
+                <Title headingLevel="h3" size="md">
+                  {t('cost_models_details.markup_or_discount')}
+                </Title>
                 <TextContent>
-                  <Text component={TextVariants.h6}>{t('cost_models_wizard.markup.sub_title')}</Text>
+                  <Text component={TextVariants.h6}>
+                    {t('cost_models_wizard.description_markup_or_discount_model')}
+                  </Text>
```
dlabrecq, line 46: This one...

src/pages/costModels/createCostModelWizard/markup.tsx
```diff
@@ -14,66 +19,100 @@ import { Form } from 'components/forms/form';
...
+                      </FormGroup>
+                    </GridItem>
+                  </Grid>
                 </Form>
               </StackItem>
+              <StackItem>
+                <div style={styles.exampleMargin}>
+                  <TextContent>
+                    <Text>{t('cost_models_details.examples.title')}</Text>
+                  </TextContent>
```
dlabrecq, line 107: this one should have an  `h` tag

src/pages/costModels/createCostModelWizard/markupWithDistribution.tsx
```diff
@@ -0,0 +1,165 @@
...
+                      </FormGroup>
+                    </GridItem>
+                  </Grid>
+                </Form>
+              </StackItem>
+              <StackItem>
+                <div style={styles.exampleMargin}>
+                  <TextContent>
+                    <Text>{t('cost_models_details.examples.title')}</Text>
+                  </TextContent>
```
dlabrecq, line 116: this one should have an  `h` tags

src/pages/costModels/createCostModelWizard/markup.tsx
ddonahue007: Added headingLevel back to the none description text fields

src/pages/costModels/createCostModelWizard/reviewWithDistribution.tsx
dlabrecq: There isn't a separate review page for price list, tho. I'm not sure I follow the logic here

ddonahue007: Yes sorry about that I should have read this a bit clearer its the review step looking at that now.

dlabrecq (approved):

PR #2055, opened 2021-08-03
Change radio buttons so they are stacked and input field is moved closer

ddonahue007:
Closes: https://issues.redhat.com/browse/COST-1738

Screen Shot 2021-08-03 at 2 51 58 PM

Screen Shot 2021-08-03 at 3 31 01 PM

### DIFF
--- a/src/pages/costModels/costModel/updateMarkupDialog.tsx
+++ b/src/pages/costModels/costModel/updateMarkupDialog.tsx
@@ -3,8 +3,6 @@ import {
   Button,
   Flex,
   FlexItem,
-  Form,
-  FormGroup,
   InputGroup,
   InputGroupText,
   List,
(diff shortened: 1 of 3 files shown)


src/pages/costModels/costModel/updateMarkupDialog.tsx
```diff
@@ -157,8 +162,8 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
                         value="false" // "+"
                         onChange={this.handleSignChange}
                       />
-                    </div>
-                    <div style={styles.radioAlign}>
+                    </FlexItem>
+                    <FlexItem style={styles.marginTopSix}>
```
dlabrecq, line 166: Please rename `marginTopSix` to something more like `radioMargin`, so we know what it's used for. If we want to change the size of the margin, then we don't need to rename the property.

src/pages/costModels/costModel/updateMarkupDialog.tsx
```diff
@@ -168,33 +173,30 @@ class UpdateMarkupModelBase extends React.Component<Props, State> {
...
-                        onKeyDown={this.handleOnKeyDown}
-                        onChange={this.handleMarkupDiscountChange}
-                        validated={this.markupValidator()}
-                      />
-                      <InputGroupText style={styles.percent}>%</InputGroupText>
-                    </InputGroup>
+                </Flex>
+                <Flex direction={{ default: 'column' }} alignSelf={{ default: 'alignSelfCenter' }}>
+                  <FormGroup fieldId="rate" helperTextInvalid={t('cost_models_wizard.markup.invalid_markup_text')}>
+                    <FlexItem style={styles.marginTopSix}>
```
dlabrecq, line 181: Instead of adding the same top margin to both radio and input text elements, add it once to their container.

src/pages/costModels/createCostModelWizard/markup.tsx
```diff
@@ -54,10 +54,10 @@ class MarkupWithDistribution extends React.Component<WithTranslation> {
               </StackItem>
               <StackItem>
                 <Form>
```
dlabrecq, line 56: Same as above, probably don't need `form` anymore?

src/pages/costModels/createCostModelWizard/markup.tsx
```diff
@@ -78,32 +78,30 @@ class MarkupWithDistribution extends React.Component<WithTranslation> {
...
-                            onKeyDown={handleOnKeyDown}
-                            onChange={handleMarkupDiscountChange}
-                            validated={markupValidator()}
-                          />
-                          <InputGroupText style={styles.percent}>%</InputGroupText>
-                        </InputGroup>
+                    </Flex>
+                    <Flex direction={{ default: 'column' }} alignSelf={{ default: 'alignSelfCenter' }}>
+                      <FormGroup fieldId="rate" helperTextInvalid={t('cost_models_wizard.markup.invalid_markup_text')}>
+                        <FlexItem style={styles.marginTopSix}>
```
dlabrecq, line 86: Same as above, add a top margin once to the parent container.

dlabrecq (changes requested):

dlabrecq (approved):

PR #2057, opened 2021-08-03
Reduce typeahead delay

dlabrecq:
During a fit-and-finish meeting, we spoke about the typeahead delay. We would like to try reducing the typeahead delay slightly.

https://issues.redhat.com/browse/COST-1742


ddonahue007 (approved):
LGTM

PR #2058, opened 2021-08-03
Page titles should use h1 tags

dlabrecq:
Ensures all page titles use h1 heading level. Also replaces strings with `TitleSizes` variants.

https://issues.redhat.com/browse/COST-1741


boaz0 (approved):

ddonahue007 (approved):
LGTM

PR #2068, opened 2021-08-06
fix for assign source button

ddonahue007:
closed: https://issues.redhat.com/browse/COST-1758

The assign sources button should be disabled until there is a change from the current config.

Screen Shot 2021-08-06 at 11 11 45 AM

With the fix we can see the button is disabled util there is change from current state/config

Screen Shot 2021-08-06 at 12 12 07 PM

Screen Shot 2021-08-06 at 12 22 49 PM

With a source previously assigned
Screen Shot 2021-08-06 at 12 25 34 PM

Screen Shot 2021-08-06 at 12 27 12 PM

### DIFF
--- a/src/pages/costModels/costModel/addSourceWizard.tsx
+++ b/src/pages/costModels/costModel/addSourceWizard.tsx
@@ -25,6 +25,7 @@ import AddSourceStep from './addSourceStep';
 
 interface AddSourcesStepState {
   checked: { [uuid: string]: { selected: boolean; meta: Provider } };
+  initialChecked: { [uuid: string]: { selected: boolean; meta: Provider } };
 }
 
 interface Props extends WithTranslation {
@@ -49,7 +50,7 @@ const sourceTypeMap = {
 };
 
 class AddSourceWizardBase extends React.Component<Props, AddSourcesStepState> {
-  public state = { checked: {} };
+  public state = { checked: {}, initialChecked: {}, initialCheckCalled: false };
   public componentDidMount() {
     const {
       costModel: { source_type },
@@ -70,6 +71,7 @@ class AddSourceWizardBase extends React.Component<Props, AddSourcesStepState> {
         };
       }, {}) as { [uuid: string]: { selected: boolean; meta: Provider } };
       this.setState({ checked: initChecked });
+      this.setState({ initialChecked: initChecked });
     }
   }
   public render() {
@@ -83,7 +85,12 @@ class AddSourceWizardBase extends React.Component<Props, AddSourcesStepState> {
         actions={[
           <Button
             key="save"
-            isDisabled={isUpdateInProgress || this.props.isLoadingSources || this.props.fetchingSourcesError !== null}
+            isDisabled={
+              this.state.checked === this.state.initialChecked ||
+              isUpdateInProgress ||
+              this.props.isLoadingSources ||
+              this.props.fetchingSourcesError !== null
+            }
             onClick={() => {
               onSave(Object.keys(this.state.checked).filter(uuid => this.state.checked[uuid].selected));
             }}
--- a/src/pages/costModels/costModel/addSourceWizard.tsx
+++ b/src/pages/costModels/costModel/addSourceWizard.tsx
@@ -50,7 +50,8 @@ const sourceTypeMap = {
 };
 
 class AddSourceWizardBase extends React.Component<Props, AddSourcesStepState> {
-  public state = { checked: {}, initialChecked: {}, initialCheckCalled: false };
+  private checkChecked = (a, b) => Object.keys(a).every(key => b[key] && a[key].selected === b[key].selected);
+  public state = { checked: {}, initialChecked: {} };
   public componentDidMount() {
     const {
       costModel: { source_type },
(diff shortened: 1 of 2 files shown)


dlabrecq (changes requested):
Tested locally and noticed that the button remains enabled after deselecting the checkbox
The select-all toggle should update the button as well

ddonahue007:
> Tested locally and noticed that the button remains enabled after deselecting the checkbox
> The select-all toggle should update the button as well


https://user-images.githubusercontent.com/57504257/128567076-9aae6b3b-9081-4ea6-abb8-9085c909a44e.mov

src/pages/costModels/costModel/addSourceWizard.tsx
```diff
@@ -49,7 +50,8 @@ const sourceTypeMap = {
 };
 
 class AddSourceWizardBase extends React.Component<Props, AddSourcesStepState> {
-  public state = { checked: {} };
+  private checkChecked = (a, b) => Object.keys(a).every(key => b[key] && a[key].selected === b[key].selected);
+  public state = { checked: {}, initialChecked: {} };
```
dlabrecq, line 54: Please keep the default `state` variables at the top of the component.

Typically, you'll find the life cycle methods appear next. The constructor (if any), followed by the `componentDidMount` and `componentDidUpdate` methods. For some reason, the `render` method is always last.

src/pages/costModels/costModel/addSourceWizard.tsx
```diff
@@ -70,6 +72,7 @@ class AddSourceWizardBase extends React.Component<Props, AddSourcesStepState> {
         };
       }, {}) as { [uuid: string]: { selected: boolean; meta: Provider } };
       this.setState({ checked: initChecked });
+      this.setState({ initialChecked: initChecked });
```
dlabrecq, line 75: We don't really care about the initial state. We only need to count the current selections using the `checked` state.

src/pages/costModels/costModel/addSourceWizard.tsx
```diff
@@ -83,7 +86,12 @@ class AddSourceWizardBase extends React.Component<Props, AddSourcesStepState> {
         actions={[
           <Button
             key="save"
-            isDisabled={isUpdateInProgress || this.props.isLoadingSources || this.props.fetchingSourcesError !== null}
+            isDisabled={
+              this.checkChecked(this.state.checked, this.state.initialChecked) ||
```
dlabrecq, line 90: Please rename `checkedChecked` to something more like `!hasSelectedItems` or `!hasSelections`.

dlabrecq (changes requested):

src/pages/costModels/costModel/addSourceWizard.tsx
```diff
@@ -49,7 +50,8 @@ const sourceTypeMap = {
 };
 
 class AddSourceWizardBase extends React.Component<Props, AddSourcesStepState> {
-  public state = { checked: {} };
+  private checkChecked = (a, b) => Object.keys(a).every(key => b[key] && a[key].selected === b[key].selected);
```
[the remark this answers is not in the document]
ddonahue007: This will not work AS if there is a preexisting item(source) that has been assign to the cost model it still shows up in the list and it shows up as selected(true). So if we test to see if an item is selected(true) and enable the button it would always be enable for any cost model that already has a source assigned to it. This is not what we want. Taking the initial value and testing it against any changes that vary from that initial value captures all test cases.  In this case we are just cycling cycling through the state objects and comparing the the "selected" lists to see if they differ.

src/pages/costModels/costModel/addSourceWizard.tsx
ddonahue007: Done

src/pages/costModels/costModel/addSourceWizard.tsx
ddonahue007: We need to compare the initial state to the current state, as the button will only be enabled when the current state differs from the initial state.

src/pages/costModels/costModel/addSourceWizard.tsx
ddonahue007: I changed this to compareSelected as that what the function is doing.

src/pages/costModels/costModel/addSourceWizard.tsx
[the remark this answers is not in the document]
ddonahue007: I am testing a new approach this morning, it still involves having an initial state as we need that, but what i am thinking is that I can capture the initial total of selected items. then we can easily have something like this

`    If (selectedCount === initialSelectedCount) {
        disableButton = true;
    }`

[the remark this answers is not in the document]
ddonahue007: I pushed back changes using counts, this should cleanup some of the comparison calls.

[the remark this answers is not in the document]
dlabrecq: Sorry if I wasn't being clear, but we can still reuse the existing `checked` state. The addSourceStep component shows the initial checked state as disabled. We simply need to check that as well, then look for at least one `selected true` item.

See https://github.com/project-koku/koku-ui/pull/2073/files

src/pages/costModels/costModel/addSourceWizard.tsx
dlabrecq: Yes, but we don't need a separate object for this. My point being that `checked` already maintains the selected state, so let's reuse it. See comments above

src/pages/costModels/costModel/addSourceWizard.tsx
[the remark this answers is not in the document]
ddonahue007: OK looked at what you had, I switched up the code base to follow that. I have to admit I missed the fact that we were already tracking the disabled in the step class.

dlabrecq (approved):

PR #2071, opened 2021-08-06
Update Travis to push to stage-beta only

dlabrecq:
Now that stage-beta is being promoted as expected, we no longer need to push to ci-beta.

https://issues.redhat.com/browse/COST-1759


ddonahue007 (approved):
LGTM

PR #2072, opened 2021-08-06
Webpack

dlabrecq:
Updates Webpack to work with both CI and stage/prod environments. Just need to close the browser in order to switch from stage to CI and vice versa, but working now.

> CI was trying to proxy to https://ci.console.redhat.com/ -- Zack

Also updated the Travis custom release script to support ci-beta and ci-stable for a bit longer. Although we still push directly from master to stage-beta.


ddonahue007 (approved):
LGTM

PR #2078, opened 2021-08-10
Eslint rules for react-intl

dlabrecq:
Added `eslint` rules to enforce the `id`, `description`, and `default-message` for `react-intl`. 

In addition, this ensures acsending sort order as previously done with `reactI18next`. For example, run the `lint:ts:fix` target to sort the messages.ts file.

https://issues.redhat.com/browse/COST-1372


.eslintrc.json
```diff
@@ -40,6 +42,9 @@
         "beforeEach": "readonly"
     },
     "rules": {
+        "formatjs/enforce-default-message": "error",
+        "formatjs/enforce-description": "error",
+        "formatjs/enforce-id": "error",
```
ddonahue007, line 47: Can we switch the sorting order on this? I under stand that this order is laying out the messages attributes in alphabetical order but it makes the file harder to read through. I think having the messages themselves in alphabetical order is a good thing but the attributes should be in the following order for easier readability.

 id: '',
description: '',
defaultMessage: '',

ddonahue007 (changes requested):

PR #2085, opened 2021-08-12
Use infrastructure.uuid property for empty states

dlabrecq:
Empty states are not being displayed correctly for "xxx filtered by OpenShift" perspectives. For this type of filter, we need to look at its associated OCP source.

A new `infrastructure.uuid` property was added to OCP sources. We can use this to match the uuid of the "AWS filtered by OpenShift" source with the OCP source. Once we have the correct OCP source, we can test its `current_month_data` and `previous_month_data` properties.

https://issues.redhat.com/browse/COST-1776

### DIFF
--- a/src/api/providers.ts
+++ b/src/api/providers.ts
@@ -30,6 +30,11 @@ export interface ProviderCostModel {
   uuid: string;
 }
 
+export interface ProviderInfrastructure {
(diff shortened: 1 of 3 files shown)


ddonahue007 (approved):
LGTM

PR #2092, opened 2021-08-17
Change function args so we can override default MessageDescriptor

ddonahue007:
Change function args so we can override default MessageDescriptor


src/utils/dateRange.ts
```diff
@@ -35,10 +36,14 @@ export function getNoDataForDateRangeString(key: string = 'no_data_for_date', of
...
   const startDate = format(startOfMonth(today), 'd');
 
-  return intlHelper(intl.formatMessage(messages.NoDataForDate, { count: getDate(today), startDate, endDate, month }));
+  return intlHelper(intl.formatMessage(key, { count: getDate(today), startDate, endDate, month }));
 }
 
-export function getForDateRangeString(value: string | number, key: string = 'for_date', offset: number = 1) {
+export function getForDateRangeString(
+  value: string | number,
+  message: MessageDescriptor = messages.ForDate,
```
dlabrecq, line 44: Kept this as `key`? Or, perhaps we should change the other `key` params to match `message`?

dlabrecq (changes requested):

dlabrecq (approved):

PR #2093, opened 2021-08-17
Added IntlProvider wrapper from Insights and default locale

dlabrecq:
Added `IntlProvider` wrapper from Insights translations pkg and default locale.


ddonahue007 (approved):
LGTM

PR #2094, opened 2021-08-18
React intl injectIntl

dlabrecq:
Updated the i18n branch to use `injectIntl` from react-intl. This allows 'EN' and 'FR' translations to be displayed, based on the user's browser locale -- mimics @redhat-cloud-services/frontend-components-translations.

Note that tests do not currently work.

Screen Shot 2021-08-17 at 10 13 11 PM


ddonahue007 (approved):
LGTM

PR #2097, opened 2021-08-18
Snake case keys

dlabrecq:
Changed message keys to use snake case for group_bys and units

Azure breakdown page, which uses `subscription_guid` as "Accounts"

Screen Shot 2021-08-18 at 7 01 40 PM

Overview page, which uses `gb_hours`, `gb_mo`, etc.

Screen Shot 2021-08-18 at 6 35 40 PM


ddonahue007 (approved):
LGTM

PR #2099, opened 2021-08-19
React intl cleanup groupby msgs

ddonahue007:
Issue: https://issues.redhat.com/browse/COST-1372

Refactor on the groupby messages in the json.messages file. This is not changing the test we see but how we use the messages themselves.


locales/data.json
```diff
@@ -81,16 +81,8 @@
...
     "GroupByTop": "DE Top {groupBy}s",
-    "GroupByTopValueAccount": "DE accounts",
-    "GroupByTopValueCluster": "DE clusters",
-    "GroupByTopValueInstanceType": "DE Instance types",
-    "GroupByTopValueNode": "DE Nodes",
-    "GroupByTopValueOrganizationalUnitID": "Organizational units",
-    "GroupByTopValuePod": "DE Pods",
-    "GroupByTopValueProject": "DE Projects",
-    "GroupByTopValueRegion": "DE Regions",
-    "GroupByTopValueService": "DE Services",
```
dlabrecq, line 92: Were these keys unused? 

Can we please make GroupByTop look similar to GroupByValues? The group_by key is not localized.

Screen Shot 2021-08-19 at 10 27 43 AM

dlabrecq (commented):
Can we please make GroupByTop look similar to GroupByValues? The group_by key is not localized.

You can see this issue best with the Azure overview.

Screen Shot 2021-08-19 at 10 27 43 AM

dlabrecq (changes requested):
This appears to be broken

Screen Shot 2021-08-19 at 10 23 20 AM

src/locales/messages.ts
```diff
@@ -495,57 +495,32 @@ export default defineMessages({
...
-  GroupByTopValueRegion: {
-    defaultMessage: 'EN Regions',
-    description: 'Regions',
-    id: 'GroupByTopValueRegion',
-  },
-  GroupByTopValueService: {
-    defaultMessage: 'EN Services',
-    description: 'Services',
-    id: 'GroupByTopValueService',
-  },
```
dlabrecq, line 542: Were these keys unused?

ddonahue007: Yes

src/utils/dateRange.ts
```diff
@@ -56,7 +56,7 @@ export function getForDateRangeString(
 
   return intlHelper(
     intl.formatMessage(message, {
-      groupByValue: intl.formatMessage(messages.GroupByValues, { value, count: 2 }),
+      groupByValue: intl.formatMessage(messages.GroupByValuesTitleCase, { value, count: 2 }),
```
dlabrecq, line 59: This isn't always a group by value. For the details pages, this is a cost.
I believe this is why the key was previously named `value`

dlabrecq (approved):

PR #2101, opened 2021-08-20
Reset API cache

dlabrecq:
This ensures that any cached API responses are reset when navigating to a new page.

https://issues.redhat.com/browse/COST-1786


ddonahue007 (approved):
LGTM

PR #2107, opened 2021-08-23
PatternFly milestone update

dlabrecq:
PatternFly milestone update.

Also had to update PatternFly icons to use esm paths, PatternFly now breaks when using a mix of esm and js paths.


ddonahue007 (approved):
LGTM

redallen (approved):

PR #2108, opened 2021-08-23
React-intl: Export modal

dlabrecq:
Updated the export modal to use react-intl

Screen Shot 2021-08-23 at 1 56 27 PM


ddonahue007 (approved):
LGTM

PR #2145, opened 2021-08-28
react-intl explorer changes

ddonahue007:
issue: COST-1372

Screen Shot 2021-08-28 at 5 25 58 PM

Screen Shot 2021-08-28 at 9 29 32 AM

Screen Shot 2021-08-28 at 5 27 01 PM


src/locales/messages.ts
```diff
@@ -314,6 +314,11 @@ export default defineMessages({
     description: 'Cost',
     id: 'Cost',
   },
+  CostExplorerTitle: {
+    defaultMessage: 'EN Cost Explorer',
+    description: 'Cost Explorer',
+    id: 'CostExplorerTitle',
+  },
```
dlabrecq, line 321: We already have `ExplorerTitle`. 

FYI, the component is named explorer, so I skip the "Cost" prefix.

src/locales/messages.ts
```diff
@@ -419,6 +424,25 @@ export default defineMessages({
...
+      'Jul {EN Jul {date}} ' +
+      'Aug {EN Aug {date}} ' +
+      'Sep {EN Sep {date}} ' +
+      'Oct {EN Oct {date}} ' +
+      'Nov {EN Nov {date}} ' +
+      'Dec {EN Dec {date}} ' +
+      'other {}}',
+    description: 'Database services cost',
+    id: 'DateMonthAbbr',
+  },
```
dlabrecq, line 445: I'm thinking we might add a new function to src/utils/dateRange.ts for this, so our abbreviations remain consistent.

Alternatively, we could add all month abbreviations to the messages file like we did in the master branch, but using select?

If we go that route, we could get rid of getAbbreviatedMonth and getLocalizedMonth from src/utils/dateRange.ts. We could also remove getDateFnsLocale from src/components/i18n/intl.ts

There appears to be about 4 messages we would need to covert?

Regardless, we should discuss which approach we want to use.

src/pages/views/explorer/explorerChart.tsx
```diff
@@ -146,42 +150,42 @@ class ExplorerChartBase extends React.Component<ExplorerChartProps> {
...
+  //       break;
+  //     case PerspectiveType.ocpCloud:
+  //       result = 'explorer.title.ocp_cloud';
+  //       break;
+  //     default:
+  //       result = undefined;
+  //       break;
+  //   }
+  //   return result;
+  // };
```
dlabrecq, line 188: Please remove

src/pages/views/explorer/explorerChart.tsx
```diff
@@ -253,7 +257,7 @@ class ExplorerChartBase extends React.Component<ExplorerChartProps> {
       <>
         <div style={styles.titleContainer}>
           <Title headingLevel="h3" size="md">
-            {t(this.getChartTitle(perspective))}
+            {intl.formatMessage(messages.PerspectiveValues, { value: perspective })}
```
dlabrecq, line 260: The chart titles are not the same as the perspective. They should contain the words "Top 5 Costliest"

```
"aws": "Amazon Web Services - Top 5 Costliest",
"aws_ocp": "Amazon Web Services filtered by OpenShift - Top 5 Costliest",
"azure": "Microsoft Azure - Top 5 Costliest",
"azure_ocp": "Microsoft Azure filtered by OpenShift - Top 5 Costliest",
"gcp": "Google Cloud Platform - Top 5 Costliest",
"gcp_ocp": "Google Cloud Platform filtered by OpenShift - Top 5 Costliest",
"ibm": "IBM Cloud - Top 5 Costliest",
"ocp": "All OpenShift cost - Top 5 Costliest",
"ocp_cloud": "All cloud filtered by OpenShift - Top 5 Costliest"
```

dlabrecq (changes requested):

src/locales/messages.ts
ddonahue007: make sense just missed that

src/locales/messages.ts
ddonahue007: I suppose, I was taking the route that we took on other select items, to eliminate Puzzling...

src/pages/views/explorer/explorerChart.tsx
ddonahue007: Done

src/utils/dateRange.ts
```diff
@@ -25,6 +25,11 @@ export function getLocalizedMonth(year, month, abbreviate: boolean = false) {
   return monthName;
 }
 
+// returns localized month name and passed in date
+export function getLocalizedAbbrMonthCost(date, month) {
+  return getLocalizedMonth(date, month, true) + ' ' + date;
+}
+
```
dlabrecq, line 32: We shouldn't concatenate strings. It would be preferable to have a localized message like: `'{month} {date}'`.

dlabrecq (approved):

PR #2147, opened 2021-08-28
Extra cost explorer requests

dlabrecq:
Cost explorer is generating extra API requests. Some query params are not valid for the current perspective because it's based on sources API requests which are still in progress.  

We need to wait until those requests have returned before attempting to generate an API request for the cost report.

https://issues.redhat.com/browse/COST-1813


ddonahue007 (approved):
LGTM

PR #2151, opened 2021-08-30
Missing "no data" labels

dlabrecq:
Added missing chart legend "no data" labels. 

Instead of the showing "no data" legend items, they should read something more like:
"Cost (no data)", "Cost forecast (no data)", and "Cost confidence (no data)"

**Before**
Screen Shot 2021-08-29 at 9 04 42 PM

**Cost and usage charts**
Screen Shot 2021-08-29 at 8 50 28 PM

Historical usage chart
Screen Shot 2021-08-29 at 8 50 12 PM


ddonahue007 (approved):
LGTM

PR #2152, opened 2021-08-30
Fixed explorer chart title

dlabrecq:
Fixed explorer chart title

**stage-stable (left-side) Vs stage-beta (right-side)**

Screen Shot 2021-08-29 at 9 20 59 PM

**After**
Screen Shot 2021-08-29 at 9 31 38 PM


ddonahue007 (approved):
LGTM

PR #2163, opened 2021-09-01
React intl markup dialog

dlabrecq:
Localized cost models markup dialog. And fixed cost models wizard message.

Screen Shot 2021-09-01 at 11 27 02 AM


ddonahue007 (approved):
LGTM

PR #2174, opened 2021-09-04
Clean up react-int testing

dlabrecq:
Cleaned up react-int testing. 

Instead of using a mock as the intl default, we can use the actual intl instance.


ddonahue007 (commented):
LGTM

PR #2176, opened 2021-09-04
Localized cost models wizard table

dlabrecq:
Localized cost models wizard table and removed unused aria-label from pagination template


ddonahue007 (commented):
LGTM

PR #2189, opened 2021-09-05
Remove react-i18next package

dlabrecq:
Remove react-i18next package and archive react-18next related files.


ddonahue007 (commented):
LGTM

PR #2194, opened 2021-09-06
Fixed Azure "filter by account" string

dlabrecq:
Fixed Azure "filter by account" string. Should be "service", not "service_name"

Screen Shot 2021-09-05 at 8 32 13 PM


ddonahue007 (commented):
LGTM

PR #2197, opened 2021-09-07
Refactored syncTranslations script to use AST

dlabrecq:
Refactored `syncTranslations` script to use Node and parse an abstract syntax tree (AST). This is a much cleaner way to add locale prefixes (e.g., FR, DE, etc.) to test messages.

Runs with the `translations:test` target.


ddonahue007 (commented):
LGTM

PR #2198, opened 2021-09-07
Pre-compile data.json messages into AST

dlabrecq:
Pre-compile `locales/data.json` messages into an abstract syntax tree (AST). 

- data.json is always the same format
- The `translations` and `translations:test` targets both generate AST
- Pre-compiling is faster 

Note that en.json, fr.json, and de.json are git ignored.
And translations.json file is not pre-compiled for readability -- only this file is sent to be translated into various locales.


ddonahue007 (commented):
LGTM

PR #2200, opened 2021-09-07
Various message fixes

dlabrecq:
Various message fixes

- Units displays an extra '{' character
- Month over month dates in details should not be abbreviated


ddonahue007 (commented):
LGTM

PR #2201, opened 2021-09-07
Filter by values should always be title-case

dlabrecq:
Filter by values should always be title case. Currently, we're showing a mix of lower-case and title-case strings.


ddonahue007 (commented):
LGTM

PR #2202, opened 2021-09-07
Localized bulk selector

dlabrecq:
Localized bulk selector in data toolbar

Screen Shot 2021-09-07 at 12 43 05 PM


ddonahue007 (commented):
LGTM

PR #2204, opened 2021-09-07
Fixed bad cost models wizard message

dlabrecq:
Fixed bad cost models wizard message with ok button of exit dialog 

Screen Shot 2021-09-07 at 2 25 19 PM


ddonahue007 (commented):
LGTM

PR #2205, opened 2021-09-07
Name and description should be title-case

dlabrecq:
Name and description should be title-case for table column headers, filters, etc.


ddonahue007 (commented):
LGTM

PR #2211, opened 2021-09-08
Cleaned up historical chart messages

dlabrecq:
Cleaned up historical chart messages.

The Cost chart's y-axis label will never have any other units like `hrs`, `GB`, etc. -- it's always a currency.
And the `HistoricalChartUnitsLabel` message is just a copy of the `Units` message.


ddonahue007 (commented):
LGTM

PR #2212, opened 2021-09-08
Currency formatting

dlabrecq:
- Refactored currency formatting for `react-intl`
- Cleaned up `formatValue` and tests
- Ensured units are provided for all `formatValue` calls


ddonahue007 (commented):
LGTM

PR #2216, opened 2021-09-09
Renamed Currency messages as CurrencyUnits

dlabrecq:
Renamed Currency messages as CurrencyUnits, so we can use Currency for a dropdown label


ddonahue007 (commented):
LGTM

PR #2219, opened 2021-09-10
Refactored to use local storage

dlabrecq:
Refactored to use local storage for storing currency units instead of cookies.


ddonahue007 (commented):
LGTM

PR #2220, opened 2021-09-10
Added localStorage utils

dlabrecq:
Added `localStorage` utils to make it easier to retrieve currency units used by APIs


ddonahue007 (commented):
LGTM

PR #2221, opened 2021-09-10
Ensure unit lookup is being used properly

dlabrecq:
Ensure the `unitlookup` function is being used properly with messages.Units


ddonahue007 (commented):
LGTM

PR #2227, opened 2021-09-16
Added formatPercentage utils

dlabrecq:
- Updated components to call `formatPercentage` instead of `formatUnits`
- Renamed `ValueFormatter` and `ValueFormatterOptions` as `Formatter` and `FormatOptions`
- Localized percent symbol shown by cost models markup

https://issues.redhat.com/browse/COST-1858

### DIFF
--- a/locales/data.json
+++ b/locales/data.json
@@ -34757,6 +34757,12 @@
         "value": " % of cost"
       }
     ],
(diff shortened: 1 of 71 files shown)


ddonahue007 (approved):
LGTM

PR #2232, opened 2021-09-24
Markup shows validation error for German locale

dlabrecq:
The cost models markup dialog currently shows a validation error for the German locale. One issue is related to the validator, which doesn't account for different decimal separators.

I refactored to format markup based on the browser's locale. In addition, I've added validation messages, similar to the rate input field. This also simplifies the modal logic a bit. 

Note: The wizard will be update via a separate PR.

**After**
chrome-capture

**Before**
Screen Shot 2021-09-24 at 12 23 34 AM


ddonahue007 (approved):
LGTM

PR #2234, opened 2021-09-24
RateInput component not localized

dlabrecq:
Localized the RateInput aria-label and placeholder.

https://issues.redhat.com/browse/COST-1893

In addition, I fixed the wizard step validator to match the markup validator. That is, the next button should not be enabled unless the markup is valid. That should include checking the number of decimal places, white spaces, etc.

**Wizard markup next step button**
chrome-capture (2)

**Add Rate placeholder**
Screen Shot 2021-09-24 at 10 37 45 AM


ddonahue007 (approved):
LGTM

PR #2235, opened 2021-09-24
Show new features in beta environment only

dlabrecq:
We want to show a couple features based only in the beta environment. Specifically, the Cost Explorer 90 days view and the currency selector in the page headers.

**Stage environment**
Screen Shot 2021-09-24 at 11 10 15 AM

**Beta environment**
Screen Shot 2021-09-24 at 11 10 50 AM


ddonahue007 (approved):
LGTM

PR #2236, opened 2021-09-24
Remove MarkupWithDistribution functions from CostModelContext

dlabrecq:
The wizard's `CostModelContext` object contains a few functions used by the `MarkupWithDistribution` step. Specifically, `handleOnKeyDown` and `markupValidator`.

Considering this step is never reused, and the functions don't set state, they may be better located in the `MarkupWithDistribution` component.

https://issues.redhat.com/browse/COST-1898


ddonahue007 (approved):
LGTM

PR #2240, opened 2021-09-29
Convert Dropdown components to use Select (1 of many)

dlabrecq:
Per the PatternFly guidelines, and after speaking with Natalie, we should be using PatternFly Select components instead of Dropdown for our overview and details pages.

This updates dropdowns for currency, "show cost as", and perspective.

chrome-capture (1)


ddonahue007 (approved):
LGTM

PR #2241, opened 2021-09-29
Update currency symbols to match ICU currencies

dlabrecq:
JavaScript supports ICU currencies, while the mocks incorrectly used ISO. This resulted in a mismatch between the currency selector symbols and the currencies formatted in the page.

https://issues.redhat.com/browse/COST-1911

chrome-capture (2)


ddonahue007 (approved):
LGTM

PR #2242, opened 2021-09-29
Add currency selection to cost models review

dlabrecq:
Adds currency selection to cost models review

https://issues.redhat.com/browse/COST-1916

Screen Shot 2021-09-29 at 2 52 21 PM


ddonahue007 (approved):
LGTM

PR #2245, opened 2021-10-01
Update axios, jest, and @redhat-cloud-services packages

dlabrecq:
Update `axios`, `jest`, and `@redhat-cloud-services` packages to help fix new vulnerability alerts. reported by Dependabot.

Note: `setImmediate` is no longer supported by `jest`, I refactored the unit tests to use `waitFor` from the @testing-library/react package.

Snapshots will never pass async `waitFor` tests, so removed a couple `toMatchSnapshot` statements -- "every call to that assertion will result in a snapshot update".


ddonahue007 (approved):
LGTM

PR #2246, opened 2021-10-04
Update css-loader

dlabrecq:
Update postcss per https://bugzilla.redhat.com/show_bug.cgi?id=1984014

The postcss v8.3.8 package is actually a dependency of css-loader, which is only a development dependency required to run the UI locally.

This is not delivered in production, but we can update css-loader from version 6.2.0 to version 6.3.0.

https://issues.redhat.com/browse/COST-1925


ddonahue007 (approved):
LGTM

PR #2247, opened 2021-10-04
Convert groupBy dropdown component to use Select

dlabrecq:
Converted `groupBy` dropdown component to use PatternFly `Select`

chrome-capture


ddonahue007 (approved):
LGTM

PR #2248, opened 2021-10-04
Convert chart comparison dropdown to PatternFly Select component

dlabrecq:
Converted the chart comparison dropdown to use PatternFly Select component

chrome-capture (1)


ddonahue007 (approved):
LGTM

PR #2251, opened 2021-10-05
Simplify cost model currency formatting

dlabrecq:
For cost models, we're currently formatting the user's input for rate and markup, based on the current locale. For the German locale, if the user enters `123.45`, the value is output as `123,45`. This works, but also unexpectedly changes characters (e.g., dot to comma).

Instead of formatting and changing the user's text input, it would be better to leave that as is. Then, simply rely on the text input validator to catch bad characters. Prior to submitting an API request, we would normalize the user's rates and markup to a decimal literal / number as expected by the API.

Likewise, when the add / edit rate and markup dialogs are launched, we can format the initial value per the current locale, but without ever touching the user's input. Again, prior to any API update, we will normalize the rates and markup to a decimal literal / number as required by the API.

Note: All the snapshots below are in the **German locale** (note the comma decimal separator) to ensure rates are formatted correctly in the UI. And un-formatted for API requests.

**Wizard**
wizard

**Edit rate**
edit rate

**Add rate**
add rate

**Edit markup**
markup


ddonahue007 (approved):
LGTM

PR #2252, opened 2021-10-06
Fixed decimal count validation

dlabrecq:
The decimal count is validated separately to show a unique error message. This particular function currently prevents that test from being performed on text inputs.

https://issues.redhat.com/browse/COST-1927
Screen Shot 2021-10-06 at 8 00 34 AM


ddonahue007 (approved):
LGTM

PR #2276, opened 2021-11-01
Simplify API requests for AWS cost savings plan

dlabrecq:
We can simplify by adding the `cost_type` param in two places. Just needed to account for the report ID used to cache requests.

https://issues.redhat.com/browse/COST-1905


ddonahue007 (approved):
LGTM

PR #2277, opened 2021-11-01
Ensure currency and cost type fetch is performed for beta env only

dlabrecq:
Ensure currency and cost type API request is performed for beta env only

https://issues.redhat.com/browse/COST-1905


dlabrecq (commented):
Pushing fix for stage-stable env

ddonahue007 (commented):
LGTM

PR #2280, opened 2021-11-02
Replace GCP project with gcp_project

dlabrecq:
For both the details page and Cost Explorer, the GCP group should read "GCP project" instead of "Project". This also applies to the "GCP filtered by OpenShift" perspective.

Instead of using the cost API's `group_by[project]` query param, we should use `group_by[gcp_project]`

https://issues.redhat.com/browse/COST-1980

**GCP details**
Screen Shot 2021-11-02 at 11 40 52 AM

**Cost Explorer**
Screen Shot 2021-11-02 at 11 40 36 AM


ddonahue007 (approved):
LGTM

PR #2281, opened 2021-11-02
Omit in-progress features from prod-beta

dlabrecq:
The platform now has an isProd flag. We can use the isProd and isBeta flag to keep new features out of prod and prod-beta. This means, users will only see our "in progress" features in stage-beta.

https://issues.redhat.com/browse/COST-2028


ddonahue007 (approved):
LGTM

PR #2282, opened 2021-11-02
AWS savings plan updates

dlabrecq:
- Ensure API query has all params for AWS savings feature 
- Ensure history.location is called after state is saved


dlabrecq (commented):
self review

PR #2285, opened 2021-11-09
Lazy generate API requests for tags

dlabrecq:
For better performance, we would like the UI to lazily generate API requests for tags.

In order to populate the "group by" dropdown, an initial API request will be made for just the tag keys. Then, we can lazily request tag key values after the user chooses to filter by a particular tag key.

https://issues.redhat.com/browse/COST-2053

https://user-images.githubusercontent.com/17481322/140975408-50186ebb-bc53-4a18-aee3-28e2258410cb.mov

### DIFF
--- a/locales/data.json
+++ b/locales/data.json
@@ -348,7 +348,15 @@
             "value": [
               {
                 "type": 0,
-                "value": "Cost by tags"
+                "value": "Back to "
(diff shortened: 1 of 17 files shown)


dlabrecq (commented):
Self review

PR #2288, opened 2021-11-15
Wrap toolbar filters with PF ToolbarToggleGroup

dlabrecq:
The toolbar isn't broken per se. The toolbar items are wrapping as expected. However, the details page toolbar wraps the filter with a PatternFly `ToolbarToggleGroup` component. 

For small windows, the PatternFly `ToolbarToggleGroup` component replaces the filter with a simple icon. When the icon is clicked, the filter is expanded onto its own line.

https://issues.redhat.com/browse/COST-2065

**Device toolbar**
after


dlabrecq (commented):
Self review

PR #2290, opened 2021-11-19
2094 pf toolbar

dlabrecq:
The PatternFly ToolbarFilter fails to save state when a "node" label is added to the user's settings list of available tags.

The debugger shows the issue may have something to do with how PatternFly tests `ToolbarChipGroup.hasOwnProperty`.

As a workaround, I'm able to provide a string to `ToolbarFilter.categoryName` instead of a `ToolbarChipGroup` object.

Screen Shot 2021-11-19 at 12 25 57 PM


dlabrecq (commented):
Self review

PR #2291, opened 2021-11-20
PatternFly ToolbarFilter error

dlabrecq:
This adds a workaround to prevent category name collisions with the PatternFly ToolbarFilter

chrome-capture


dlabrecq (commented):
Self review

PR #2292, opened 2021-11-23
Update category prefix

dlabrecq:
Make category prefix longer to prevent key collisions


dlabrecq (commented):
Self review

PR #2294, opened 2021-11-23
PatternFly ToolbarFilter

dlabrecq:
This simplifies the previous workaround to prevent category name collisions with the PatternFly ToolbarFilter.

Note that the tag key value `Select` component already uses an existing tag prefix, so we can take advantage of that instead of adding a new prefix to the category `ToolbarFilter`.


dlabrecq (commented):
Self review

PR #2297, opened 2021-12-01
Lazy generate API requests for tags

dlabrecq:
This ensures the tag API request is not generated while already in progress.

https://issues.redhat.com/browse/COST-2053


dlabrecq (commented):
Self review

PR #2298, opened 2021-12-02
Remove isBeta flag from AWS saving plan feature

dlabrecq:
- Remove `isBeta` flag from AWS saving plan feature 
- Sync with the user's preferred choice from the settings page

https://issues.redhat.com/browse/COST-2108


dlabrecq (commented):
Self review

PR #2300, opened 2021-12-03
Cost type options appear bold in breakdown header

dlabrecq:
Moved the cost type out of the PatternFly Title component's body.

**Before**
Screen Shot 2021-12-03 at 9 12 06 AM

**After**
Screen Shot 2021-12-03 at 9 26 54 AM


dlabrecq (commented):
Self review

PR #2301, opened 2021-12-08
Can't assign multiple sources to a cost model across paginated pages

dlabrecq:
The addSourceWizard.tsx component resets the selected list of items whenever new sources are loaded; for example, when paginating between table items. If we omit that, selections will be saved across pages.

https://issues.redhat.com/browse/COST-2097

chrome-capture


dlabrecq (commented):
Self review

PR #2303, opened 2021-12-14
Replace type-ahead with PatternFly's Text Input Group component

dlabrecq:
Replace the existing `Select` type-ahead with PatternFly's new `Text Input Group` component.

https://issues.redhat.com/browse/COST-2159

chrome-capture


dlabrecq (commented):
Self review

PR #2304, opened 2021-12-14
Modified the "add sources" dialog to retain the applied sources

dlabrecq:
Modified the "add sources" dialog to retain the applied sources. Currently, anything previously assigned is replaced, unless everything is deselected first.

https://issues.redhat.com/browse/COST-2097

chrome-capture


dlabrecq (commented):
Self review

PR #2307, opened 2021-12-14
Add limit param to the sources API request

dlabrecq:
Noticed an issue with the overview and details pages, using the insights-qa user. We recently added 15 sources to test a bug, but now those pages display an empty state indicating we're processing data.

For the overview and details pages, the UI looks at the current_month_data property returned by the sources API. By default, the API only returns 10 items, and none of those sources have data. We need to look at all the sources.

To resolve this issue, we should add the `limit=100` param to the sources API request. For example:

api/cost-management/v1/sources/?type=OCP&limit=100


dlabrecq (commented):
Self review

PR #2308, opened 2021-12-14
Adjusted the default perspective for the overview infrastructure tab

dlabrecq:
Adjusted the default perspective for the overview infrastructure tab. This will help ensure the default overview tab has data available.


dlabrecq (commented):
Self review

PR #2309, opened 2021-12-14
Clean up default perspective

dlabrecq:
Clean up default perspective for the overview infrastructure tab. We just need to change the default perspective for "all cloud".


dlabrecq (commented):
Self review

PR #2310, opened 2021-12-15
Simplify provider utils

dlabrecq:
Simplify provider 'has data' utils by removing fetch status


dlabrecq (commented):
Self review

PR #2311, opened 2021-12-15
Clean up explorer default perspective utils

dlabrecq:
Clean up explorer default perspective utils


dlabrecq (commented):
Self review

PR #2313, opened 2022-01-13
Replace cost-type and currency requests with account-settings API

dlabrecq:
In order to reduce the number of requests made by the UI, we would like to omit the cost-type and currency API requests. The UI already localizes those strings and they're not likely to change.

This will replace cost-type and currency requests with the account-settings API.

https://issues.redhat.com/browse/COST-2201


dlabrecq:
Self review

PR #2314, opened 2022-01-17
Request sources API without type param

dlabrecq:
Modified UI to use a single sources API request instead of using the type param.

https://issues.redhat.com/browse/COST-2202


dlabrecq (commented):
Self review

PR #2315, opened 2022-01-24
Enable GCP on OpenShift

dlabrecq:
In order to enable GCP on OpenShift, the following UI changes are required.

For GCP on OpenShift, we should use `group_by[project]` and the `resource-types/gcp-projects` API to filter.
And for GCP, we continue to use `group_by[gcp_project]` and the `resource-types/gcp-gcp-projects` (new) API to filter.


dlabrecq (commented):
Self review

PR #2316, opened 2022-01-24
Updated tab key navigation for type-ahead

dlabrecq:
The PatternFly Text Input Group component, used for type-ahead has issues regarding keyboard navigation. 

When navigating items with the tab key, focus should move back to the input, before moving to another element in the page. This will allow the user to edit the text input and / or tab to the clear button.

https://issues.redhat.com/browse/COST-2203


dlabrecq (commented):
Self review

PR #2318, opened 2022-01-26
Add limit param to metrics API request

dlabrecq:
The metrics API request is not picking up newly added units. There are now 11 metrics, but the API returns 10 by default.

We need to add a `limit` parameter to metrics API request, so we can display all units.

https://issues.redhat.com/browse/COST-2262


dlabrecq (commented):
Self review

PR #2320, opened 2022-01-27
Refactor user access and sources API requests

dlabrecq:
Refactor API requests to ensure user access and sources are fetched prior to rendering each page. This will help to ensure RBAC and source updates are reflected in the UI.

https://issues.redhat.com/browse/COST-2268


dlabrecq (commented):
Self review

PR #2322, opened 2022-01-28
Effective-usage metric should include units

dlabrecq:
When creating rates, "Effective-usage" is presented as an option, but it should include units, similar to other localized options.

https://issues.redhat.com/browse/COST-2277


dlabrecq (commented):
Self review

PR #2324, opened 2022-02-03
Cleaned up component names to match future export component

dlabrecq:
Renamed *View components as *Content to match PatternFly drawer component


dlabrecq (commented):
Self review

PR #2325, opened 2022-02-03
Add "All exports" link

dlabrecq:
Add "All exports" link to open export drawer per mocks. The contents of the drawer will be added later, as APIs become available. For now, this will only be visible in stage beta.

https://issues.redhat.com/browse/COST-2305

chrome-capture (3)


dlabrecq (commented):
Self review

PR #2327, opened 2022-02-10
Use chrome HTML template for local development.

Hyperkid123:
### Changes
- update webpack-dev-server
- use chrome HTML template in development environment; New version of webpack dev server broke ESI tags mutation due to closed response headers.


dlabrecq (commented):
The solution seems ok, but I wasn't able to get the UI running with stage-beta.

Hyperkid123:
> The solution seems ok, but I wasn't able to get the UI running with stage-beta (via our "start" target).

Right, I forgot to remove the original template override. It should be working now, can you verify?

> Curious why the templates are used only with isProduction?

Because in a dev environment, the HTML middleware would override the chrome template that is duplicated before the dev server is turned on. We don't want to use that template anymore. This change is now in stage beta but not yet in production.

dlabrecq (approved):
This appears to be working now. Thank you, @Hyperkid123

PR #2329, opened 2022-02-10
Enable 90 day UI view in stable stage/prod

dlabrecq:
Enable 90 day UI view in stable stage/prod

https://issues.redhat.com/browse/COST-2210


dlabrecq (commented):
Self review

PR #2335, opened 2022-02-16

dlabrecq:
Add content to "All exports" drawer.

- Added drawer description and initial toolbar for filter and pagination
- Added initial table to "All exports" drawer
- Added status icons to all exports drawer

Screen Shot 2022-02-16 at 4 20 56 PM


dlabrecq (commented):
Self review

PR #2336, opened 2022-02-17

dlabrecq:

- Fixed download and delete icon colors
- Fixed bottom pagination margins

Screen Shot 2022-02-17 at 9 39 55 AM


dlabrecq (commented):
Self review

PR #2339, opened 2022-02-22
Updated export dialog to match "all exports" mocks

dlabrecq:
Updated export dialog to match "all exports" mocks.

Note: changes are shown for beta-only

**OpenShift Details**
Screen Shot 2022-02-22 at 5 33 06 PM

**Cost Explorer**
Screen Shot 2022-02-23 at 7 50 40 AM


dlabrecq (commented):
Self review

PR #2344, opened 2022-02-24
Bullet chart requests shown as range for no capacity

dlabrecq:
When showing bullet charts without capacity, the requests are shown in grey instead of blue. For example, when OCP is grouped by project instead of cluster.

In this case, requests are shown as the bullet chart's "qualitative range". To be consistent, requests and usage should be shown together as the "primary segmented measure".

https://issues.redhat.com/browse/COST-2363

**Before (grouped by project)**
no capacity

**After (grouped by project)**
Screen Shot 2022-02-24 at 8 30 08 AM


dlabrecq (commented):
Self review

PR #2345, opened 2022-02-24
Remove unused groupBy instance_type

dlabrecq:
Remove unused groupBy instance_type support and clean up unused i18n messages.

Grouping by instance_type may have been used long ago, with the overview summary card tabs. However, it appears there are no longer any current use cases.

For many reports, groupBy instance_type is not a valid parameter. It seems that only AWS and Azure support it.

https://issues.redhat.com/browse/COST-2361


dlabrecq (commented):
Self review

PR #2348, opened 2022-02-25

dlabrecq:

- Added validation for export name
- All inputs in export dialog should be required

**Name validation**
Screen Shot 2022-02-25 at 10 05 18 AM


dlabrecq (commented):
Self review

PR #2356, opened 2022-03-07
Fix remote container initialization.

Hyperkid123:
Recent versions of webpack-dev-server are unable to run remote containers locally without injecting the container entry into HTML if a custom `splitChunks` configuration is defined.

The custom splint chunks config is ignored by webpack if a module is a part of a remote container but the dev server is still picking it up. Dropping it fixes the remote container initialization which is required


dlabrecq (approved):
Tested locally and the UI is working again. Thank you for all the help!

PR #2358, opened 2022-03-09
Update how "Show cost as" value is cached

dlabrecq:
Update how "Show cost as" value is cached

Testing:

1. Login via incognito window - to start with a clean slate
2. Navigate to account settings
3. Update the default "Show cost as" value
4. Navigate to AWS details and see the default value
5. Select a new "Show cost as" value (same page)
6. Refresh the page and confirm the new value has been applied
7. There should now be a cost_type query param in the page URL
8. Navigate to the Overview page and AWS perspective
9. Confirm the new value has been applied
10. Navigate to the Cost Explorer page and AWS perspective
11. Confirm the new value has been applied
12. Refresh the page
13. Confirm the default value is shown

https://issues.redhat.com/browse/COST-2451

### DIFF
--- a/src/api/queries/query.ts
+++ b/src/api/queries/query.ts
@@ -23,6 +23,7 @@ export interface Filters {
 
 export interface Query {
   cost_type?: any;
+  currency?: any;
   dateRange?: any;
   end_date?: any;
   filter?: any;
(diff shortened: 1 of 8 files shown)


dlabrecq (commented):
Self review

PR #2359, opened 2022-03-16
Document how to setup /etc/hosts

dlabrecq:
Document how to setup /etc/hosts.

At one time, we relied on the insights-proxy and its setup instructions. We no longer use that proxy, but we still need to setup the hosts file for local development.

https://issues.redhat.com/browse/COST-2476


dlabrecq (commented):
Self review

PR #2363, opened 2022-04-01
Connect currency selection to cost model API

dlabrecq:
The currency selector was already added via COST-1856 and available via the stage-beta env. However, we still need to provide the user's chosen currency to the cost-model API.

This change allows cost-models to be created in different currencies. In addition, rates can be edited and added in the cost model currency. Cost model currencies are always localized per the browser's locale.

https://issues.redhat.com/browse/COST-1907

**Cost model price list:**
Screen Shot 2022-04-01 at 2 44 56 PM

### DIFF
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -206,7 +206,7 @@ export const isPercentageFormatValid = (value: string) => {
 };
 
 // Some locales have a comma decimal separator (e.g., "1.234,56" in German is "1,234.56" in USD).
-// This function normalizes a given currency or percentage for APIs.
+// This function normalizes a given currency or percentage for APIs where USD decimal formatting is expected.
 export const unFormat = (value: string) => {
   if (!value) {
     return value;
--- a/src/utils/format.ts
+++ b/src/utils/format.ts
@@ -206,7 +206,7 @@ export const isPercentageFormatValid = (value: string) => {
 };
 
 // Some locales have a comma decimal separator (e.g., "1.234,56" in German is "1,234.56" in USD).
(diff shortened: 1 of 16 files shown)


src/utils/format.ts
```diff
@@ -206,7 +206,7 @@ export const isPercentageFormatValid = (value: string) => {
 };
 
 // Some locales have a comma decimal separator (e.g., "1.234,56" in German is "1,234.56" in USD).
-// This function normalizes a given currency or percentage for APIs.
+// This function normalizes a given currency or percentage for APIs where USD decimal format is expected.
 export const unFormat = (value: string) => {
```
gitdallas, line 210: i wonder if the name of this should be changed to be more descriptive rather than having to add comments wherever it is being used.. maybe `normalizeToUsdDecimalFormat` even

dlabrecq (commented):
Self review

src/utils/format.ts
dlabrecq: Sorry I missed this. Currency support is a work in progress, and will consider your suggestion for the next PR. For now, I'd like to keep the comments.

dlabrecq: Wanted to follow up on this. I gave the function name some thought, but would like to keep it and add comments to better describe how it's used. I forgot it's not only used for APIs.

1. Some locales us a comma as the decimal separator (e.g., "1.234,56" in German), which must be replaced for APIs where USD decimal format is expected.

2. The group separator (e.g., "1,234.56" in USD) must also be removed when formatting currencies and percentages to display in the browser's locale.

PR #2364, opened 2022-04-01
Update Travis to use node 16

dlabrecq:
It appears that the user-event package requires a newer version of node. 
We should bump the Travis build to use node 16.

https://issues.redhat.com/browse/COST-2526


gitdallas (approved):

dlabrecq (commented):
LGTM

PR #2367, opened 2022-04-05
GCP breakdown not showing all cards

dlabrecq:
The GCP breakdown is missing a few cards when grouping by GCP Project; services, regions, etc. 

https://issues.redhat.com/browse/COST-2533

Screen Shot 2022-04-05 at 11 59 36 AM


dlabrecq (commented):
Self review

PR #2369, opened 2022-04-06
Increase number of group_by options shown

dlabrecq:
Increase number of group_by options shown for org units and tag keys. Also cleaned up some unused filters.

https://issues.redhat.com/browse/COST-2536

Screen Shot 2022-04-06 at 7 52 47 PM


gitdallas (approved):

dlabrecq (commented):
Merging

PR #2375, opened 2022-04-20
Add testing library and jest dom linters, fix tests to work with those

gitdallas:
also fixed a typo in readme about call to update manifest and got the selector to default to appending menu to parent, but the wizard uses will still append to body

https://issues.redhat.com/browse/COST-2573


src/pages/costModels/createCostModelWizard/generalInformation.tsx
```diff
@@ -139,6 +140,7 @@ class GeneralInformation extends React.Component<GeneralInformationProps> {
                     <Selector
                       label={messages.Currency}
                       direction={SelectDirection.up}
+                      appendMenuTo={() => document.body}
```
dlabrecq, line 143: Doesn't this break accessibility?

src/pages/costModels/costModel/updateRateModel.test.tsx
```diff
@@ -259,127 +257,131 @@ describe('update-rate', () => {
...
+      userEvent.click(screen.getByLabelText('measurement label'));
     });
-    userEvent.click(getAllByRole('option')[1]);
-    expect(getByText(regExp(messages.Save)).closest('button').disabled).toBeFalsy();
+    userEvent.click(screen.getAllByRole('option')[1]);
+    expect(saveButton.getAttribute('disabled')).toBeNull();
 
     await waitFor(() => {
-      userEvent.click(document.querySelector(qr.measurement));
+      userEvent.click(screen.getByLabelText('measurement label'));
```
dlabrecq, line 286: Should these strings come from our messages? I see a mix of both below

src/pages/costModels/costModel/sourcesToolbar.tsx
```diff
@@ -80,7 +80,7 @@ export const SourcesToolbar: React.SFC<SourcesToolbarProps> = ({
           </ToolbarItem>
         </ToolbarToggleGroup>
         <ToolbarItem>
-          <ReadOnlyTooltip isDisabled={actionButtonProps.isDisabled}>
+          <ReadOnlyTooltip isDisabled={true}>
```
dlabrecq, line 83: Do we want to change this, it may be used dynamically elsewhere?

dlabrecq (commented):
Can you please create a Jira issue for this? Just need a little explanation as to what's changing.

I see new messages, new aria labels, etc. Are you changing all cost model tests to use `userEvent` and `screen`?

I see new messages for place holder text, new aria labels, etc.

src/pages/costModels/costModel/sourcesToolbar.tsx
dlabrecq: It appears that the `isDisabled` property is based on write permissions, so probably don't want to hard code this to always be disabled. I believe the intention is for `ReadOnlyTooltip` to omit the tooltip when a user doesn't have RBAC privileges.

Note that the `cost-demo` user has admin privileges in RBAC, so you may not see this change unless you try another user. We have multiple users to test different privileges (e.g., cost-cloud, cost-openshift, etc.).

src/pages/costModels/createCostModelWizard/generalInformation.tsx
gitdallas: It does, this is what is needed to make it work with the wizard though. Previously I was appending to body all the time, this change is to make it only do so in this case which is within the wizard.

Up to you on waiting.

src/pages/costModels/costModel/updateRateModel.test.tsx
gitdallas: testing-library does not like selectors that come from the deconstructed render. i added the label as a way to select the element, i'm open to other options though.

src/pages/costModels/costModel/sourcesToolbar.tsx
gitdallas: Oh sorry, I was testing something out and forgot to revert

src/pages/costModels/createCostModelWizard/generalInformation.tsx
dlabrecq: Let's wait until PatternFly has a chance to investigate on their end. Ideally, they will provide a fix that allows us to append to the parent.

I know appending to the body works with these tests. However, after meeting with the PatternFly folks, appending to the body isn't recommended for various reasons. Mainly because it effects keyboard navigation.

After further testing, I can confirm that keyboard navigation doesn't work for the Select in the wizard's general information step.

src/pages/costModels/costModel/updateRateModel.test.tsx
dlabrecq: I don't feel strongly about it, but would prefer if the strings came from messages. Ideally, strings would be maintained in one place.

gitdallas: i discovered a `toggleAriaLabel` prop in the PF component that will allow me to put the aria label directly on the Select button, so I can target by that rather than having to deal with hidden label/labelby or testId+querySelector... I've also used the messages for strings.

src/pages/costModels/createCostModelWizard/generalInformation.tsx
gitdallas: this update is a step in that direction. i moved all but the wizard things to append to parent rather than body (this would include "add rate" and "edit rate") - so it is better than existing in that regard.

tests also look better.

dlabrecq (approved):
Thanks for creating https://issues.redhat.com/browse/COST-2617. We can merge this PR now.

PR #2378, opened 2022-04-20
Modify DataListCheck to use isChecked

dlabrecq:
Modify DataListCheck to use isChecked after latest PatternFly milestone breakage

Screen Shot 2022-04-20 at 3 48 46 PM

https://issues.redhat.com/browse/COST-2571


gitdallas (approved):

gitdallas:
tested on main and this doesn't break anything there so I don't think we need to merge

PR #2380, opened 2022-05-10
update select menu appends

gitdallas:
https://issues.redhat.com/browse/COST-2617

ended up using inline in cases where the menu for sure fits between edges of modal and doesn't go over header/footer of wizard with the new height restriction (scrollable), and parent for cases where it goes off the edge of the modal. both ways work with keyboard and a11y (being close in dom).

the 350 max height currently only affects the currency select, and gives it most of the usable space between wizard header/footer while cutting off at half an item so it's more obvious that scrolling is possible.

### DIFF
--- a/src/pages/costModels/components/inputs/selector.tsx
+++ b/src/pages/costModels/components/inputs/selector.tsx
@@ -19,6 +19,7 @@ interface SelectorFormGroupOwnProps {
   label?: MessageDescriptor | string;
   appendMenuTo?: HTMLElement | 'parent' | 'inline' | (() => HTMLElement);
   toggleAriaLabel?: string;
+  maxHeight?: string | number;
   placeholderText?: string;
   direction?: SelectDirection.up | SelectDirection.down;
   options: {
@@ -51,6 +52,7 @@ const SelectorBase: React.FunctionComponent<SelectorProps> = ({
   id,
   intl = defaultIntl, // Default required for testing
   toggleAriaLabel,
+  maxHeight,
   placeholderText,
   direction = SelectDirection.down,
   isInvalid = false,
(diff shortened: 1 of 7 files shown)


src/pages/costModels/components/inputs/selector.tsx
```diff
@@ -94,6 +94,7 @@ const SelectorBase: React.FunctionComponent<SelectorProps> = ({
     >
       <Select
         id={id}
+        maxHeight={350}
```
dlabrecq, line 97: There is `maxHeight` and other variables stored in wizard.styles. Can we add this value there?

dlabrecq (commented):
Looks good, just one small nit.

I ran this locally, but can you please add a screenshot to the PR. Just in case Natalie or QE wants to review?

gitdallas:
https://user-images.githubusercontent.com/5322142/168086854-c154ea38-881d-4438-93de-eb02cc1b4d7f.mov

nlcwong:
LGTM

src/pages/costModels/components/inputs/selector.tsx
gitdallas: I was applying the maxHeight to any use of the select dialog. Right now wizard styles are being used by the generalInformation component only. They are also always being used as real styles (passed into `style` attributes on basic HTML elements), this value is a custom PF prop that gets handled by PF rather than as an actual style attribute.

If desired, I could create a maxHeight prop on our Selector component that we then pass into the PF Select component.. and just pass it to our Selector component from generalInformation with a `maxHeight={styles.selector.maxHeight`.

dlabrecq: Yes, `maxHeight={styles.selector.maxHeight}` is what I was looking for. We typically place all style properties, like height in a `styles` file -- even though PatternFly may treat this one differently. Easier to find properties and / or reuse them when located together.

I agree that a `maxHeight` prop would be best, so we can use it only when necessary.

gitdallas:
added component ids in for qa testers, added aria labels on the generalInfo select toggles, created a maxHeight property in our selector to pass to PF select and setting to 350 for general info items

dlabrecq (approved):
LGTM

PR #2384, opened 2022-05-18
Dependabot vulnerability alert

dlabrecq:
GitHub's dependabot shows a "Prototype Pollution in async" alert. This dependency is inherited from webpack-dev-server.

https://issues.redhat.com/browse/COST-2644


dlabrecq (commented):
This pkg update will remove the `async` dependency

PR #2388, opened 2022-06-03
Sync "All filtered by Openshift" network filters

dlabrecq:
The "All filtered by Openshift" view should include "GCP" network filters. We currently only include AWS and Azure.

https://issues.redhat.com/browse/COST-2691


dlabrecq (commented):
Note that https://github.com/project-koku/koku-ui/pull/2387 is a prereq for this PR.

gitdallas (approved):

PR #2390, opened 2022-06-03
fix issue with sorting on measurement in rate table

gitdallas:
https://issues.redhat.com/browse/COST-2702

when creating the data for the cells, this column includes the measurement_unit, however it is not included currently in the filter.. added that with this PR

### DIFF
--- a/src/pages/costModels/components/rateTable.tsx
+++ b/src/pages/costModels/components/rateTable.tsx
@@ -39,7 +39,7 @@ const RateTableBase: React.SFC<RateTableProps> = ({ actions, intl = defaultIntl,
         sortBy.index === 1
           ? (r: Rate) => r.metric.label_metric
           : sortBy.index === 2
-          ? (r: Rate) => r.metric.label_measurement
+          ? (r: Rate) => r.metric.label_measurement + r.metric.label_measurement_unit
           : () => '';
       return compareBy(r1, r2, sortBy.direction, projection);
     })


src/pages/costModels/components/rateTable.tsx
```diff
@@ -39,7 +39,7 @@ const RateTableBase: React.SFC<RateTableProps> = ({ actions, intl = defaultIntl,
         sortBy.index === 1
           ? (r: Rate) => r.metric.label_metric
           : sortBy.index === 2
-          ? (r: Rate) => r.metric.label_measurement
+          ? (r: Rate) => r.metric.label_measurement + r.metric.label_measurement_unit
```
dlabrecq, line 42: I have a couple issues with the existing sort, not necessarily your fix.

1. This isn't a localized sort and will break for other locales, eventually.

Although this appears to address the immediate problem, we're not sorting on the formatted strings displayed in the UI. For example, we're adding spaces, parens, etc. via localized strings.

2. The code is performing a client-side sort, which is a problem with paginated tables.
 
FYI, we cannot perform a client-side sort properly while items remain server-side. For example, if a user is viewing 10 items per page and there are multiple pages of data. We can only sort on the 10 items in view for the current page. Any items located on another page would be excluded from that sort.

That said; after looking at the cost-models API, I suspect rates are NOT paginated. It appears that only individual cost models are paginated. We should confirm with the backend team, but we can likely remove the pagination controls and stick with a localized, client-side sort here.

Can we please:

1. Sort localized strings
2. Remove the pagination controls from the price list table
3. Confirm with the backend team that rates are not paginated

If rates are paginated, then please create a new Jira issue and request that the cost-models API support server-side sorting.

dlabrecq (changes requested):

src/pages/costModels/components/rateTable.tsx
dlabrecq: Thought this string was localized, but should be. We should never hard code characters like spaces and parens like this.

`${tier.metric.label_measurement} (${tier.metric.label_measurement_unit})`,

gitdallas: Ah, so we stumbled across some bigger issue. I'll update the jira with this info

PR #2391, opened 2022-06-06
AWS cost type dropdown state not updating

dlabrecq:
The AWS cost type dropdown state is not updating. This change ensures the dropdown state is updated as expected.

https://issues.redhat.com/browse/COST-2672

chrome-capture-2022-5-6


gitdallas (approved):
lgtm, but wonder if the cost type util should be named as such to avoid confusion with the costType component and property

PR #2392, opened 2022-06-07
update rate table to use translations and filter correctly, remove pagination

gitdallas:
https://issues.redhat.com/browse/COST-2702


src/pages/costModels/costModel/priceListTable.tsx
```diff
@@ -306,29 +276,6 @@ class PriceListTable extends React.Component<Props, State> {
...
-                            onPerPageSelect={(_evt, perPage) =>
-                              this.setState({
-                                pagination: { page: 1, perPage },
-                              })
-                            }
-                            variant="bottom"
-                          />
-                        </ToolbarItem>
-                      </ToolbarContent>
-                    </Toolbar>
```
dlabrecq, line 331: Trying to find where this was used in the UI? For example, the cost model wizard still has pagination for rates.

src/pages/costModels/costModel/priceListTable.tsx
```diff
@@ -152,15 +139,11 @@ class PriceListTable extends React.Component<Props, State> {
         />
         <WithPriceListSearch initialFilters={{ primary: 'metrics', metrics: [], measurements: [] }}>
           {({ search, setSearch, onRemove, onSelect, onClearAll }) => {
-            const from = (this.state.pagination.page - 1) * this.state.pagination.perPage;
-            const to = this.state.pagination.page * this.state.pagination.perPage;
```
dlabrecq, line 156: Did you happen to try adding a bunch of rates to see if pagination was working? 

After looking at the code that was removed, it looks that pagination was also a client-side implementation. Thinking we should run a quick test before we remove it.

dlabrecq (changes requested):

src/pages/costModels/costModel/priceListTable.tsx
gitdallas: ah, I'll take a look

src/pages/costModels/costModel/priceListTable.tsx
gitdallas: would you rather have a client side pagination vs no pagination?

gitdallas: it looks like pagination works client side... with the exception of sorting not including pages you aren't currently viewing

src/pages/costModels/costModel/priceListTable.tsx
gitdallas: it looks like the cost model wizard has it's own implementation..  https://github.com/project-koku/koku-ui/blob/main/src/pages/costModels/createCostModelWizard/priceListTable.tsx#L205

src/pages/costModels/components/rateForm/rateForm.tsx
```diff
@@ -59,19 +59,18 @@ const RateFormBase: React.FunctionComponent<RateFormProps> = ({
...
-    });
-    return label ? label : m;
+    const units = intl.formatMessage(messages.Units, { units: unitsLookupKey(u) }) || u;
+    return (
+      intl.formatMessage(messages.MeasurementValues, {
+        value: m.toLowerCase().replace('-', '_'),
+        units,
+        count: 2,
+      }) || m
+    );
```
dlabrecq, line 73: I was thinking `unitsLookupKey` would work for the value as well. Would something like this sudo code work?
    
```
    const value = unitsLookupKey(u);
    const units = intl.formatMessage(messages.Units, { units: value }) || u;
    return (
      intl.formatMessage(messages.MeasurementValues, {
        value,
        units,
        count: 2,
      }) || m
    );
```

src/utils/format.ts
```diff
@@ -238,12 +238,16 @@ export const unitsLookupKey = (units): string => {
...
+    case 'cluser_month':
+    case 'pvc_month':
     case 'gibibyte_month':
     case 'hour':
     case 'hrs':
     case 'tag_mo':
     case 'vm_hours':
       return lookup;
+    case 'gb_month':
+      return 'gb_mo';
```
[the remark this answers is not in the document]
dlabrecq: That's fine

src/pages/costModels/costModel/priceListTable.tsx
dlabrecq: Since pagination is functional, we should keep it.

I can see the sort doesn't work quite as expected. Although, I believe we may have a workaround below.

In order for a sort to work with pagination, it needs to be applied before rates are paginated. Thinking we could move the sort function from `rateTable` to `priceListTable`, before we `slice` rates.  Then, add an `onSort` prop to `rateTable` so we can pass table events to `priceListTable`.

Does that sound doable to you?

dlabrecq (approved):
LGTM

PR #2399, opened 2022-06-29
fix issue with warnings in logs on datalistcheck change

gitdallas:
Austin Sullivan says he believes this is a bug in DataListCheck and that I can use the `checked` property instead.

https://issues.redhat.com/browse/COST-2770

### DIFF
--- a/src/pages/views/details/components/columnManagement/columnManagementModal.tsx
+++ b/src/pages/views/details/components/columnManagement/columnManagementModal.tsx
@@ -156,7 +156,7 @@ export class ColumnManagementModalBase extends React.Component<ColumnManagementM
               <DataListItemRow>
                 <DataListCheck
                   aria-labelledby={option.value}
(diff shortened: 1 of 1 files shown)


dlabrecq (changes requested):
We had used the `checked` prop previously and it broke with this milestone build:
https://github.com/project-koku/koku-ui/pull/2378. 

At the time, the PatternFly team told us that the correct / expected solution is to use `isChecked`. 

It appears to be working in the UI, but per Slack conversation, it still sounds like a workaround. I'd prefer to wait for the fix below in the next PF milestone.
https://github.com/patternfly/patternfly-react/pull/7635

gitdallas:
Yea, I believe they ended up making the checked work back then... and apparently created this issue when implementing isChecked.

I am okay waiting as it is just a console warning anyway and we won't have to do anything when the fix comes out.

PR #2417, opened 2022-08-03
update pf and insights deps

gitdallas:
https://issues.redhat.com/browse/COST-2854


gitdallas:
the problem with tests right now is that there are ids being generated in what is making it to the snap... i'm going to look into that yet

dlabrecq (approved):

PR #2424, opened 2022-08-18
update test/jest dependencies

gitdallas:
https://issues.redhat.com/browse/COST-2827

- will have to wait until react 18 before updating testing-library/react to 13+
- had to change userEvent calls as userEvent 14+ uses async for all apis
- had to make a couple updates to jest config and transformTS after updating jest


dlabrecq (approved):
LGTM

PR #2427, opened 2022-08-24
Add option to start UI with local cloud services config route

dlabrecq:
This adds a script option to start the UI with a local Insights cloud services config route. The script generates the prompt below.

"Do you want to use local cloud services config? y/N"

This will option will set the `useLocalCloudServicesConfig` env variable. That is, instead of manually editing our `webpack.config` file to apply the route below.

```
if (useLocalCloudServicesConfig) {
  routes['/beta/config'] = {
    host: `http://$/{localhost}:8889`,
  };
}
```

https://issues.redhat.com/browse/COST-2910

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -19,9 +19,6 @@ const distDir = path.resolve(__dirname, './dist/');
 const betaEnv = process.env.BETA_ENV;
 const nodeEnv = process.env.NODE_ENV;
 
-// Set `true` for testing cloud-services-config https://github.com/RedHatInsights/cloud-services-config#testing-your-changes-locally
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -20,7 +20,7 @@ const betaEnv = process.env.BETA_ENV;
 const nodeEnv = process.env.NODE_ENV;
 
 // Set `true` for testing cloud-services-config https://github.com/RedHatInsights/cloud-services-config#testing-your-changes-locally
```
gitdallas, line 22: I think this comment should be removed as we won't set here... but possibly move the link to the prompt message or to a console output at the end of the build if the option is selected?

dlabrecq: I moved the comment alongside the route code.

gitdallas (approved):

PR #2444, opened 2022-09-12
Feat/auto dep updates

gitdallas:
story: https://issues.redhat.com/browse/COST-2936

moved this into the normal PR workflow with conditions as it seems to work well in that same order with some additions.

limiting PRs created by dependabot with max open PRs to 2 for now while we make sure things work well on the main repo

can see what a dependabot + manifest update PR should look like here: https://github.com/gitdallas/koku-ui/pull/6

and that a non-dependabot PR isn't affected here: https://github.com/gitdallas/koku-ui/pull/7

i add a "dependency-bot-wip" label to dependabot PRs, which my new actions remove after updating manifest.. although it shouldn't take long.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -6,7 +6,10 @@ on:
     branches: [ main, master, prod-beta, prod-stable, stage-beta, stage-stable ]
 env:
   BRANCH: ${{ github.base_ref }}
-
+permissions:
+  pull-requests: write
+  packages: write
+  contents: write
 jobs:
   build:
     name: koku-ui build
@@ -17,6 +20,11 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
+      - name: Checkout PR
+        if: ${{github.actor == 'dependabot[bot]' }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh pr checkout ${{ github.event.pull_request.number }}
       - name: Setup Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v1
         with:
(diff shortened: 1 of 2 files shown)


gitdallas:
looks like i need to do something different about removing my label:

[Error: .github#L1](https://github.com/project-koku/koku-ui/commit/835327817a3f2aa1b55d9e9ca1d141f701eba65f#annotation_4584352455)
actions-ecosystem/action-remove-labels@v1 is not allowed to be used in project-koku/koku-ui.

.github/workflows/pull_request.yml
```diff
@@ -42,6 +50,25 @@ jobs:
...
+        if: ${{github.actor == 'dependabot[bot]' }}
+        run: node scripts/createManifest.js
+      - name: Push koku-manifest changes
+        if: ${{github.actor == 'dependabot[bot]' }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git add .
+          git commit --message "Update koku-manifest file"
+          git push
```
dlabrecq, line 71: Could we please move the new actions to a separate "manifest" workflow and change pull_request to run like so?

```
# Trigger when the manifest workflow succeeded
on:
  workflow_run:
    workflows: ["manifest build"]
    types:
      - completed
```

I'm thinking a separate workflow would be cleaner because its a dependabot only feature. It may also be easier to debug and/or disable if needed?

Note that we're migrating to the Insights container build pipeline shortly, so we won't use "pull_request" workflow to build PRs. However, it's possible that we could continue to using a separate "manifest" workflow, while disabling just the "pull_request" workflow.

dlabrecq (commented):
Have you been able to test this with the new container build? Currently, github actions and "ci.ext.devshift.net" are both building our PRs. I suspect the container build will fail, unless we can ensure the "manifest" workflow runs first?

.github/workflows/pull_request.yml
```diff
@@ -17,6 +20,11 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
+      - name: Checkout PR
+        if: ${{github.actor == 'dependabot[bot]' }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh pr checkout ${{ github.event.pull_request.number }}
```
dlabrecq, line 27: Could we please move the new actions to a separate "manifest" workflow and change pull_request.yml to run like so?

```
# Trigger when the manifest workflow succeeded
on:
  workflow_run:
    workflows: ["manifest build"]
    types:
      - completed
```

I'm thinking separate workflows would be cleaner because it's a dependabot only feature and easier to disable the "pull_request" workflow, eventually.

Note that we're migrating to the Insights container build pipeline by the end of Q4, so we won't likely use the "pull_request" workflow to build PRs. However, it's possible that we could continue to using a separate "manifest" workflow, while disabling the "pull_request" workflow only.

.github/workflows/pull_request.yml
dlabrecq, line 71: If we push to the same PR, will this build exit? I'm assuming that this PR update generates a second build?

gitdallas: it looks like the actions only ran once on this PR (didn't run on the koku-manifest commit) https://github.com/gitdallas/koku-ui/pull/6

PR #2456, opened 2022-09-14
add npm check updates and a script target for it

gitdallas:
okay to assign to https://issues.redhat.com/secure/RapidBoard.jspa?rapidView=5662&view=detail&selectedIssue=COST-2936 ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "scripts": {
     "build": "yarn build:prod",
     "build:prod": "yarn clean && node --max_old_space_size=8192 node_modules/.bin/webpack --mode=production",
+    "check:updates": "yarn npm-check-updates",
     "clean": "rimraf dist/",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "build": "yarn build:prod",
     "build:prod": "yarn clean && node --max_old_space_size=8192 node_modules/.bin/webpack --mode=production",
+    "check:updates": "yarn npm-check-updates",
```
dlabrecq, line 13: Does this need to be `npx npm-check-updates`?

gitdallas: no, it's working with yarn
image

dlabrecq (approved):

package.json
dlabrecq: Ok, I mentioned it because we have other script targets that use `npx`

PR #2464, opened 2022-09-16
replace human-date with date-fns

gitdallas:
[COST-2938](https://issues.redhat.com/browse/COST-2938)

### DIFF
--- a/src/routes/costModels/costModelsDetails/utils/table.tsx
+++ b/src/routes/costModels/costModelsDetails/utils/table.tsx
@@ -2,7 +2,9 @@ import { Bullseye } from '@patternfly/react-core';
 import { IAction, ICell, SortByDirection } from '@patternfly/react-table';
 import { Unavailable } from '@redhat-cloud-services/frontend-components/Unavailable';
 import { CostModel } from 'api/costModels';
-import { formatDistanceToNow } from 'date-fns';
+import { intl } from 'components/i18n';
+import { format as formatDate } from 'date-fns';
+import messages from 'locales/messages';
 import React from 'react';
 import { Link } from 'react-router-dom';
 import { EmptyFilterState } from 'routes/components/state/emptyFilterState/emptyFilterState';
@@ -40,6 +42,12 @@ export function getRowsByStateName(stateName: string, data: any) {
     ];
   }
   return data.map((item: CostModel) => {
+    const dateTime = intl.formatMessage(messages.dateTime, {
+      dateTime: formatDate(
+        new Date(new Date(item.updated_timestamp).toISOString().replace(/z$/i, '')),
+        'dd LLL yyyy kk:mm'
+      ),
+    });
     return {
       cells: [
         {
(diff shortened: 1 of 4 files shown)


src/routes/costModels/costModelsDetails/utils/table.tsx
```diff
@@ -48,7 +48,7 @@ export function getRowsByStateName(stateName: string, data: any) {
         item.description,
         item.source_type,
         item.sources.length.toString(),
-        relativeTime(item.updated_timestamp),
+        formatDistanceToNow(new Date(item.updated_timestamp), { addSuffix: true }),
```
dlabrecq, line 51: There are a couple issues with the `formatDistanceToNow` function

1. We were showing the last change in days (91 days ago), now it's shown in months (3 months ago)
2. We need to localize the string

Didn't realize this string wasn't localized before, but it appears that `formatDistanceToNow` accepts a locale option. You can get the current locale, using:

`import { getLocale } from 'components/i18n';`

dlabrecq (changes requested):

src/routes/costModels/costModelsDetails/utils/table.tsx
gitdallas: I wonder if we'll need a custom function to use days vs months if we want to keep that format, it doesn't look like it is supported with this package.

gitdallas: might be able to do this i guess https://date-fns.org/v2.29.3/docs/intlFormatDistance

it has a `unit` option.. so it could always be days?

src/routes/costModels/costModelsDetails/utils/table.tsx
```diff
@@ -40,6 +42,12 @@ export function getRowsByStateName(stateName: string, data: any) {
     ];
   }
   return data.map((item: CostModel) => {
+    const dateTime = intl.formatMessage(messages.dateTime, {
+      dateTime: formatDate(
+        new Date(new Date(item.updated_timestamp).toISOString().replace(/z$/i, '')),
```
gitdallas, line 47: new Date will convert the time to local because of the Z at the end and would then give a "-5 GMT" for example... as we are adding UTC timestamp ourselves, I'm removing the Z to prevent this

dlabrecq (changes requested):

src/routes/costModels/costModelsDetails/utils/table.tsx
gitdallas: that's actually a separate dependency, is that okay?

dlabrecq: I think using date-fns-tz would be ok. When I created the issue, it was because we were using two completely unrelated packages for dates. 

The date-fns-tz package is an extension of date-fns, but probably not a full replacement. We use a few more functions that date-fns-tz does not appear to support; endOfMonth, getDate, getYear, startOfMonth, etc.

dlabrecq: I would be ok using date-fns-tz along with date-fns. When I created the issue, it was because we were using two unrelated packages for dates. The date-fns-tz package is an extension of date-fns, so seems like a better fit.

The date-fns-tz package probably is not a full replacement for date-fns. We're using a few more functions, which date-fns-tz does not appear to support; endOfMonth, getDate, getMonth, getYear, startOfMonth, etc.

dlabrecq (approved):

PR #2477, opened 2022-09-20
Aria updates for Overview page

gitdallas:
https://issues.redhat.com/browse/COST-3026

I've fixed everything that makes sense on the overview page. Went from 73 to to 51.

Outstanding issues are either:

-  Insights (outside CostManagement section)
- "sufficient color contrast" which is due to the chart labels because
-- "Element content is too short to determine if it is actual text content" (for the bottom because "1", etc are "too short")
-- "Unable to determine contrast ratio" (for the left side which we have hidden?)
-- "Element's background color could not be determined because element contains an image node" for the line legends

The ones within cost management are all "for review" which means it's not necessarily an error, but something they want to bring to our attention in case it is.


src/routes/views/components/charts/trendChart/trendChart.tsx
```diff
@@ -321,7 +322,7 @@ class TrendChartBase extends React.Component<TrendChartProps, State> {
         gutter={20}
         height={25}
         itemsPerRow={legendItemsPerRow}
-        name="legend"
+        name={chartName + '-legend'}
```
dlabrecq, line 325: Changing this name may break the interactive legend events. That is, when hovering over the legend, the associated data series is highlighted. Clicking on the legend item will hide / show the series.

dlabrecq: This name must match line 308 

https://github.com/project-koku/koku-ui/blob/0f570ecb28670fde733cdf917a7496f769a9060f/src/routes/views/components/charts/trendChart/trendChart.tsx#L308

```
     legendName: 'legend',
```

src/routes/views/components/charts/trendChart/trendChart.tsx
```diff
@@ -310,7 +311,7 @@ class TrendChartBase extends React.Component<TrendChartProps, State> {
     return result;
   }
 
-  private getLegend = () => {
+  private getLegend = (chartName = 'chart') => {
     const { legendItemsPerRow } = this.props;
```
dlabrecq, line 315: This could be simplified by adding `chartName` to the destructured props on line 315.
```
    const { chartName, legendItemsPerRow } = this.props;
```

src/routes/views/components/charts/usageChart/usageChart.tsx
```diff
@@ -256,7 +257,7 @@ class UsageChartBase extends React.Component<UsageChartProps, State> {
         height={25}
         gutter={20}
         itemsPerRow={itemsPerRow}
-        name="legend"
+        name={chartName + '-legend'}
```
dlabrecq, line 260: same as above

src/routes/views/overview/components/dashboardWidgetBase.tsx
```diff
@@ -262,12 +263,14 @@ class DashboardWidgetBase extends React.Component<DashboardWidgetProps> {
           formatOptions={trend.formatOptions}
           formatter={chartFormatter || formatUnits}
           height={height}
+          intl={intl}
```
dlabrecq, line 266: We shouldn't pass `intl` as a prop. We've refactored the UI to remove this. Instead, we should wrap the component with `injectIntl`.

That said, I don't believe `ReportSummaryTrend` needs to localize itself, since it just contains a chart.

Both `TrendChart` and `DailyTrendChart` are already wrapped by `injectIntl`, so I'm curious why intl was used here?

src/routes/views/overview/components/dashboardWidgetBase.tsx
```diff
@@ -473,6 +479,7 @@ class DashboardWidgetBase extends React.Component<DashboardWidgetProps> {
         previousRequestData={previousRequestData}
         previousUsageData={previousUsageData}
         title={title}
+        chartName={chartName}
```
dlabrecq, line 482: This could be simplified as chartName={trend.chartName}

src/store/dashboard/awsDashboard/awsDashboardWidgets.ts
```diff
@@ -38,6 +38,7 @@ export const computeWidget: AwsDashboardWidget = {
     computedReportItem: ComputedReportItemType.usage,
     computedReportItemValue: ComputedReportItemValueType.total,
     titleKey: messages.dashboardDailyUsageComparison,
+    chartName: 'dashboardDailyUsageComparison',
```
dlabrecq, line 41: Would you mind prefixing the chart names with "aws"; for example, "awsComputeChart"

src/store/dashboard/awsDashboard/awsDashboardWidgets.ts
```diff
@@ -94,6 +96,7 @@ export const databaseWidget: AwsDashboardWidget = {
     computedReportItem: ComputedReportItemType.cost,
     computedReportItemValue: ComputedReportItemValueType.total,
     titleKey: messages.dashboardCumulativeCostComparison,
+    chartName: 'dashboardCumulativeCostComparisonDatabase',
```
dlabrecq, line 99: Would you mind prefixing the chart names with "aws"; for example, "awsDatabaseChart"

src/store/dashboard/awsDashboard/awsDashboardWidgets.ts
```diff
@@ -143,6 +147,7 @@ export const storageWidget: AwsDashboardWidget = {
     computedReportItem: ComputedReportItemType.usage,
     computedReportItemValue: ComputedReportItemValueType.total,
     titleKey: messages.dashboardDailyUsageComparison,
+    chartName: 'dashboardDailyUsageComparisonStorage',
```
dlabrecq, line 150: Would you mind prefixing the chart names with "aws"; for example, "awsStorageChart"

src/store/dashboard/awsDashboard/awsDashboardWidgets.ts
```diff
@@ -120,6 +123,7 @@ export const networkWidget: AwsDashboardWidget = {
     computedReportItem: ComputedReportItemType.cost,
     computedReportItemValue: ComputedReportItemValueType.total,
     titleKey: messages.dashboardCumulativeCostComparison,
+    chartName: 'dashboardCumulativeCostComparisonNetwork',
```
dlabrecq, line 126: Would you mind prefixing the chart names with "aws"; for example, "awsNetworkChart"

src/store/dashboard/awsDashboard/awsDashboardWidgets.ts
```diff
@@ -66,6 +67,7 @@ export const costSummaryWidget: AwsDashboardWidget = {
     computedReportItemValue: ComputedReportItemValueType.total,
     dailyTitleKey: messages.awsDailyCostTrendTitle,
     titleKey: messages.awsCostTrendTitle,
+    chartName: 'awsCostTrend',
```
dlabrecq, line 70: Would you mind prefixing the chart names with "aws"; for example, "awsCostChart"

src/store/dashboard/awsOcpDashboard/awsOcpDashboardWidgets.ts
```diff
@@ -138,6 +142,7 @@ export const storageWidget: AwsOcpDashboardWidget = {
     computedReportItem: ComputedReportItemType.usage,
     computedReportItemValue: ComputedReportItemValueType.total,
     titleKey: messages.dashboardDailyUsageComparison,
+    chartName: 'dashboardDailyUsageComparisonStorage',
```
dlabrecq, line 145: Same as above

dlabrecq (changes requested):

src/routes/views/components/charts/trendChart/trendChart.tsx
gitdallas: good catch

src/routes/views/overview/components/dashboardWidgetBase.tsx
gitdallas: It's weird... I never made any related changes to the component... but I'm getting the following without intl

> TS2741: Property 'intl' is missing in type '{ adjustContainerHeight: boolean; containerHeight: number; currentData: ChartDatum[]; forecastData: any; forecastConeData: any; formatOptions: FormatOptions; ... 8 more ...; chartName: string; }' but required in type 'WrappedComponentProps<"intl">'.

src/routes/views/components/charts/trendChart/trendChart.tsx
```diff
@@ -49,6 +49,7 @@ interface TrendChartOwnProps {
   showSupplementaryLabel?: boolean; // Show supplementary cost labels
   showUsageLegendLabel?: boolean; // The cost legend label is shown by default
   title?: string;
+  chartName: string;
```
gitdallas, line 52: @dlabrecq - apparently this is the culprit for `intl` being required prop for `ReportSummaryComponet`... making `chartName` optional doesn't matter.. any idea why this would be?

src/routes/views/overview/components/dashboardWidgetBase.tsx
gitdallas: `intl` being required seems to be due to adding a property to `TrendChartOwnProps`, any idea why this would cause this?

src/routes/views/components/charts/trendChart/trendChart.tsx
```diff
@@ -299,19 +300,20 @@ class TrendChartBase extends React.Component<TrendChartProps, State> {
...
   // Returns onMouseOver, onMouseOut, and onClick events for the interactive legend
   private getEvents() {
+    const { chartName } = this.props;
     const { hiddenSeries, series } = this.state;
 
     const result = getInteractiveLegendEvents({
       chartNames: getChartNames(series),
       isHidden: index => isSeriesHidden(hiddenSeries, index),
-      legendName: 'legend',
+      legendName: chartName ? chartName + '-legend' : 'legend',
```
dlabrecq, line 309: I've been testing this locally and the highlighting doesn't work as expected. Some legend items remain greyed out. 

Probably a bug with the chart's `getInteractiveLegendEvents` function. I suspect "legend" may be hard coded somewhere?

Looking at the PatternFly code now to see if I can quickly fix that

dlabrecq: Created a PR to fix the legend issue in PatternFly.
https://github.com/patternfly/patternfly-react/pull/8035

src/routes/views/overview/components/dashboardWidgetBase.tsx
dlabrecq: I modified the code locally and was able to remove the `i18n` prop by making `chartName` optional.

src/routes/views/components/charts/trendChart/trendChart.tsx
```diff
@@ -321,7 +323,7 @@ class TrendChartBase extends React.Component<TrendChartProps, State> {
         gutter={20}
         height={25}
         itemsPerRow={legendItemsPerRow}
-        name="legend"
+        name={chartName ? chartName + '-legend' : 'legend'}
```
dlabrecq, line 326: You can simplify this as

name={`${chartName}-legend`}

It's a required property in the widget trend props.

src/routes/views/overview/components/dashboardWidgetBase.tsx
dlabrecq: Actually it's because `DailyTrendChart` does not support the `chartName` prop. 

`ReportSummaryComponent` is providing `chartName` to either chart `ReportSummaryDailyTrend` (i.e., the `DailyTrendChart`) or `ReportSummaryTrend` (i.e., the `TrendChart`).

So, both charts must support the same required prop or you must make it optional

src/store/dashboard/awsDashboard/awsDashboardWidgets.ts
```diff
@@ -38,6 +38,7 @@ export const computeWidget: AwsDashboardWidget = {
     computedReportItem: ComputedReportItemType.usage,
     computedReportItemValue: ComputedReportItemValueType.total,
     titleKey: messages.dashboardDailyUsageComparison,
+    chartName: 'awsComputeChart',
```
dlabrecq, line 41: Nit. Can you please make the props alphabetical?

src/routes/views/overview/components/dashboardWidgetBase.tsx
gitdallas: ah... wonder why making it optional didn't work for me yesterday... maybe i just didn't wait long enough for the ide to update.. still seems weird that it would complain about intl rather than chartname.

dlabrecq (approved):

PR #2498, opened 2022-09-27
Remove chart feature for showing infrastructure costs

dlabrecq:
At one time, we had a feature to show infrastructure costs in the overview and historical charts. There is a lot of code to support data; current, previous, forecast, forecast cone. In addition to a lot of messages, tooltips, etc. It doesn't appear likely that the functionality will be restored, so we can clean this up.

https://issues.redhat.com/browse/COST-3059


dlabrecq (commented):
Restarted container build

PR #2500, opened 2022-09-27
Clarify when cumulative data is used

dlabrecq:
In `chartDayumUtils`, `ChartType.daily` is used to represent cumulative data, while `ChartType.rolling` is the default. This is confusing because data is transformed based on a chart view, which can be cumulative for both daily and monthly data. For example, in the HCS UI, we're now showing cumulative monthly data.

That said, this PR refactors the code to use `DatumType.cumulative` and `DatumType.rolling`, which can be applied to both daily and monthly data. This will make it clear when cumulative data is being used.

https://issues.redhat.com/browse/COST-3063


dlabrecq (commented):
Self

gitdallas (approved):

PR #2501, opened 2022-09-27
Export dialog button remains disabled

dlabrecq:
The submit button in the export dialog remains disabled. It's not obvious to users why the button is disabled, but it prevents exporting data.

https://issues.redhat.com/browse/COST-3064


dlabrecq (commented):
Self

gitdallas (approved):

PR #2512, opened 2022-09-29
Revert eslint to previous version to fix linter

dlabrecq:
This keeps breaking my IDE


dlabrecq (commented):
Keep reverting this :(

PR #2513, opened 2022-09-29
Negative filtering

dlabrecq:
Adds the excludes select component to support negative filtering

https://issues.redhat.com/browse/COST-2773

chrome-capture-2022-8-29


dlabrecq (commented):
Self

gitdallas (approved):

PR #2520, opened 2022-10-01
Ensure Unleash toggles are based on user ID

dlabrecq:
Want to ensure Uleash feature toggles are enabled per account basis in the stage env.

For example, cost-demo, cost-cloud, etc. should be able to see in-progress features, but insights-qa should only see completed features. The insights-qa user should only test what users will see in prod.

https://issues.redhat.com/browse/COST-3079


gitdallas (approved):

dlabrecq (commented):
Self

PR #2529, opened 2022-10-05
Connect currency selection to cost report APIs

dlabrecq:
This adds the user's currency selection to our report API requests.

Note: Currency support has not been added to the forecast APIs, which should be available soon. And OCI has no currency support.

https://issues.redhat.com/browse/COST-1906
https://issues.redhat.com/browse/COST-2517 (forecasts)

### DIFF
--- a/src/routes/components/currency/currency.tsx
+++ b/src/routes/components/currency/currency.tsx
@@ -7,11 +7,12 @@ import React from 'react';
 import { injectIntl, WrappedComponentProps } from 'react-intl';
 import { connect } from 'react-redux';
 import { createMapStateToProps } from 'store/common';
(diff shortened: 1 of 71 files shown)


dlabrecq (commented):
Self

gitdallas (approved):
lgtm, but wonder if we could reuse some? for example,  `handleCurrencySelected` is implemented the same way in 9 places.

dlabrecq:
I started to move `handleCurrencySelected`, but may finish that in a separate PR? This touches another function which is used in a lot of places. I ended up refactoring 7 functions just for AWS details, so it's a fairly good size PR itself.

PR #2535, opened 2022-10-05
OCI missing units

dlabrecq:
Added units for BYTES_MS, GB_MS, and MS

https://issues.redhat.com/browse/COST-3086

oci


dlabrecq (commented):
Updated unit test

PR #2536, opened 2022-10-05
GCP compute card shows negative zero value

dlabrecq:
Modified the code to not show a negative zero as the card’s hero number. It’s possible you could see -100, but not -0

https://issues.redhat.com/browse/COST-3087

Screen Shot 2022-10-05 at 3 46 19 PM


dlabrecq (commented):
Self

gitdallas (approved):

PR #2537, opened 2022-10-06
Feature flags component generates state warning

dlabrecq:
The feature flags component (for Unleash) occasionally generates a "React state update on an unmounted component" warning. In addition, I see multiple requests generated for the user-access, sources, and settings API. 

This change refactors the code a bit to ensure neither will happen.


dlabrecq (commented):
Self

gitdallas (approved):

PR #2546, opened 2022-10-06
Move common functions to utils

dlabrecq:
Some new functions have been added to the details pages for currency support. Want to refactor this a bit and move common functions to the utils area. For example, `handleCurrencySelection` is something that is used in multiple components.


dlabrecq (commented):
Self

gitdallas (approved):

PR #2549, opened 2022-10-07
Remove store reset and add useLayoutEffect

dlabrecq:
Remove store reset and add useLayoutEffect


dlabrecq (commented):
Self

PR #2550, opened 2022-10-07
Restore hasFlags

dlabrecq:
Restore hasFlags and add messages to debug


dlabrecq (commented):
Self

PR #2555, opened 2022-10-10
Remove flagsReady hook

dlabrecq:
Remove Unleash's flagsReady hook. 

This does not appear to be doing anything as it's always true. And don't want to block the UI from rendering, considering the UI will update when Redux dispatches flag events.


dlabrecq (commented):
Self

PR #2564, opened 2022-10-11
fix js errors on form errors within cost model wizard

gitdallas:
https://issues.redhat.com/browse/COST-3077

### DIFF
--- a/src/routes/costModels/createCostModelWizard/steps.tsx
+++ b/src/routes/costModels/createCostModelWizard/steps.tsx
@@ -1,20 +1,20 @@
+import { intl } from 'components/i18n';
 import messages from 'locales/messages';
-import { MessageDescriptor } from 'react-intl';
 import { countDecimals, isPercentageFormatValid } from 'utils/format';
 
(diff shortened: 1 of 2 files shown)


src/routes/costModels/createCostModelWizard/steps.tsx
```diff
@@ -1,20 +1,20 @@
...
   }
   if (name.length > 100) {
-    return messages.costModelsInfoTooLong;
+    return intl.formatMessage(messages.costModelsInfoTooLong);
   }
   return null;
 };
 
-export const descriptionErrors = (description: string): MessageDescriptor | null => {
+export const descriptionErrors = (description: string): string | null => {
```
dlabrecq, line 15: I believe the intention here was to return a `MessageDescriptor`, but call `formatMessage` where we already have access to the `i18n` object in `generalInformation.tsx`.

See https://github.com/project-koku/koku-ui/blob/main/src/routes/costModels/createCostModelWizard/generalInformation.tsx#L112

dlabrecq (changes requested):

dlabrecq (approved):

PR #2566, opened 2022-10-11
Bump Redux dependency

dlabrecq:
Bump Redux dependency

https://issues.redhat.com/browse/COST-2906


dlabrecq (commented):
Self

src/utils/omit.ts
```diff
@@ -0,0 +1,4 @@
+// Based on https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html
+export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
```
gitdallas, line 2: Omit is part of typescript 3.5+, I don't think we need this.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html#the-omit-helper-type

src/utils/omit.ts
```diff
@@ -0,0 +1,4 @@
...
+
+export type DistributiveOmit<T, K extends keyof T> = T extends unknown ? Omit<T, K> : never;
```
gitdallas, line 4: not being used?

gitdallas (changes requested):

src/utils/omit.ts
dlabrecq: Updated

gitdallas (approved):

PR #2577, opened 2022-10-13
Remove start_date & end_date from cost explorer URL

dlabrecq:
The Cost explorer URL currently contains a `start_date` and `end_date`; however, these query params are not used directly as the page is based on the `dateRange` param. (These params are only used for the API request itself.) Want to clean this up and remove `start_date` & `end_date` from Cost explorer URL.

https://issues.redhat.com/browse/COST-3149


gitdallas (approved):

dlabrecq (commented):
Self

PR #2582, opened 2022-10-15
Fix export file dates

dlabrecq:
- Fix export file dates
- Cleanup date util file names

https://issues.redhat.com/browse/COST-3149


dlabrecq (commented):
Self

PR #2614, opened 2022-10-21
Revert @redhat-cloud-services/frontend-components-config pkg

dlabrecq:
Dependabot updated the UI to @redhat-cloud-services/frontend-components-config v4.6.23.

The build is successful, but with a lot of warnings. Unfortunately, the UI will not run and we must revert back to v4.6.22.

Created https://issues.redhat.com/browse/RHCLOUD-22594 to track


dlabrecq (commented):
Self

PR #2615, opened 2022-10-21
Remove unused eslint-plugin-cypress pkg

dlabrecq:
The eslint-plugin-cypress package was added because @redhat-cloud-services/frontend-components-config pkg had a missing dependency. The Insights team recently removed this dependency, so we should remove it as well.

https://issues.redhat.com/browse/COST-3183


dlabrecq (commented):
Self

gitdallas (approved):

PR #2619, opened 2022-10-21
Fix/3026 explorer a11y props

gitdallas:
https://issues.redhat.com/browse/COST-3026

@dlabrecq  I noticed that there is an extra update in the translations that isn't in the messages that is getting updated with this. Seems to be related back to https://github.com/gitdallas/koku-ui/commit/9a6105c21b5940f162bb0165bf1c7719d9563f63 - perhaps you made an extra change and forgot to run translation?


dlabrecq (approved):
Thanks for updating the translations.

Just need to fix a lint issue

PR #2625, opened 2022-10-24
chore(deps):3189 remove ncu dep

gitdallas:
as we are using npx for ncu calls now - we no longer need it as a dependency


dlabrecq (changes requested):
I'd like to keep this as a dev dependency. I'm not sure why we would remove the `npm-check-updates` dependency, unless you installed packages globally?

gitdallas:
I suppose I was just thinking: why should we keep a dependency when npx is able to pull the executable without having it as a dependency that we have to manage?

https://www.section.io/engineering-education/npm-vs-npx/#:~:text=many%20added%20features.-,Using%20NPX,and%20then%20run%20the%20package.

> With NPX, you can run and execute packages without having to install them locally or globally.
> 
> When running NPM executables with NPX, if a package is installed, NPX will search for the package binaries (either locally or globally) and then run the package.
> 
> If the package was not previously installed, NPX will not install the package in your system; instead, it will create a temporary cache that will hold the package binaries. Once the execution is over, NPX will remove the installed cache binaries from the system.

gitdallas:
As for formatjs... it looks like we wouldn't be able to remove that dependency because we using it for imports for `MessageDescriptor` in several places.

dlabrecq (approved):
I didn't realize that npx will install the dependency if necessary, so this should be fine. One less package to maintain

PR #2629, opened 2022-10-25
Added consistent-type-imports rule

dlabrecq:
- Added @typescript-eslint/consistent-type-imports eslint rule
- More missing export types per @typescript-eslint/consistent-type-exports eslint rule


dlabrecq (commented):
"@typescript-eslint/consistent-type-exports" is temporarily disabled until tests can be omitted properly

gitdallas (approved):
curious why this lint catches the enums and others but the codemod didn't, but it's not needed at runtime so that's fine.

PR #2648, opened 2022-10-28
Reset currency selection when the default is updated via settings page

dlabrecq:
This change resets the currency selection when the default currency is updated via settings page. This removes two query params from the browser URL; `currency` and `cost_type`. Using the existing properties stored via local storage only, so we don't override the default.

FYI, there is an edge case where a user sets the default currency after visiting a Cost Management page. Then, the user navigates back to the UI, but our currency dropdown remains unchanged. In this scenario, UI's currency selection had already been cached for the session. The user won't see the dropdown match the default currency until they login again.

https://issues.redhat.com/browse/COST-3207


dlabrecq (commented):
self

gitdallas (approved):

PR #2661, opened 2022-11-03
Migrate details pages to use PatternFly composable table

dlabrecq:
Migrating the remaining details pages to use the newer PatternFly composable version. The legacy table will be deprecated for the upcoming PatternFly breaking change release.

https://issues.redhat.com/browse/COST-2349

Screen Shot 2022-11-03 at 3 53 37 PM


dlabrecq (commented):
Updated to remove unused translations and fixed data labels

src/locales/messages.ts
```diff
@@ -1275,7 +1265,7 @@ export default defineMessages({
       'ocp {OpenShift details table} ' +
       'other {}}',
     description: 'Breakdown back to page titles',
-    id: 'breakdownBackToTitles',
+    id: 'dataTableAriaLabel',
```
gitdallas, line 1268: Should the description be updated?

gitdallas (approved):
Looks good. Nice to see some simplification and code reduction in some places. A little nit:

src/locales/messages.ts
dlabrecq: Yes, good catch

PR #2672, opened 2022-11-08
Migrate cost explorer page to use PatternFly composable table

dlabrecq:
- Migrate cost explorer page to use PatternFly composable table
- Add "nowrap" style to details tables

https://issues.redhat.com/browse/COST-3231

**Explorer table:**
Screen Shot 2022-11-07 at 11 20 39 PM


dlabrecq (commented):
self

PR #2695, opened 2022-11-15
Rename Unleash flag

dlabrecq:
Rename Unleash flag from "default projects" to "platform costs"


dlabrecq (commented):
Self

PR #2701, opened 2022-11-17
Reset pagination and row selections

dlabrecq:
Reset pagination and row selections when "sum platform costs" is toggled


dlabrecq (commented):
self

PR #2702, opened 2022-11-17
Export should include AWS cost type

dlabrecq:
The export feature should include AWS cost type when generating API requests for the AWS details page and Cost Explorer


dlabrecq (commented):
Self

PR #2703, opened 2022-11-18
Export should include AWS cost type (2 of 2)

dlabrecq:
This ensures the export feature uses the same query params used to generate API reports for the details table.


dlabrecq (commented):
Self

PR #2706, opened 2022-11-18
Cleanup query string names

dlabrecq:
Renamed query and queryString as reportQueryString, tagQueryString, etc.
Renamed getReportId, getExportId, etc. as getFetchId


dlabrecq (commented):
Self

PR #2707, opened 2022-11-18
Disabled details table kebab

dlabrecq:
- Disabled details table kebab for 'no-project' projects
- Fixed wrong query type
- Renamed query as userAccessQuery
- Refactored details table row selection state


dlabrecq (commented):
Self

PR #2712, opened 2022-11-21
Add platform costs support for breakdown page

dlabrecq:
Added platform costs support for breakdown page

Note that the cost-demo user does not have volumes for "platform costs", so the "Volume" chart shows zero values.

Screen Shot 2022-11-21 at 12 26 14 PM


dlabrecq (commented):
self

PR #2715, opened 2022-11-21
Simplify platform costs impl

dlabrecq:
Simplify platform costs using multiple group_by's


dlabrecq (commented):
Self review

PR #2716, opened 2022-11-22
OCP cluster breakdown card layout

dlabrecq:
Ensure cpu, memory, and volume cards appear on right side of page for OCP cluster breakdown.

Screen Shot 2022-11-21 at 7 09 34 PM


dlabrecq (commented):
self

PR #2724, opened 2022-11-28
Clear date picker's end date

dlabrecq:
Clearing the date picker's end date per QE / UX feedback. If the end date becomes invalid, after choosing a new start date, we will clear the end date.

https://issues.redhat.com/browse/COST-3271

chrome-capture-2022-10-28 (1)


dlabrecq (commented):
Self review

PR #2727, opened 2022-11-29
Fix query for details breakdown

dlabrecq:
Fix query for details breakdown return link, so page can be restored.

https://github.com/project-koku/koku-ui/pull/2711


dlabrecq (commented):
self

PR #2735, opened 2022-11-30
Simplify platform costs feature

dlabrecq:
Refactored breakdown page to simplify the platform costs feature. 

FYI, the "category" query param is only used to restore the details page. It doesn't indicate if the current project is a "platform" cost. For that, we must look at the group_by's provided to the breakdown page.

In the screenshots below, the "platform" project shows new features as expected. However, the "analitics" project should not display them.

**Example: "platform"**
platform

**Example: "analitics"**
analytics

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "xstate": "^4.34.0"
   },
   "devDependencies": {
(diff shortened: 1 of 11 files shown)


dlabrecq (commented):
self

PR #2736, opened 2022-11-30
Show unallocated (aka idle) costs in the details pages

dlabrecq:
The APIs automatically provide unallocated costs in the details table, but we need to disable those links to the breakdown page.

https://issues.redhat.com/browse/COST-3248

Screen Shot 2022-11-30 at 3 28 03 PM


dlabrecq (commented):
self

PR #2741, opened 2022-12-01
Refactor platform costs to use category filter

dlabrecq:
Refactor platform costs to use category filter instead of kube- and openshift- filters


dlabrecq (commented):
self

PR #2742, opened 2022-12-01
Ensure unallocated links are disabled.

dlabrecq:
Ensure unallocated links are disabled.

Note that the classification prop is only available for category=platform

https://issues.redhat.com/browse/COST-3248


dlabrecq (commented):
self

PR #2743, opened 2022-12-01
Ensure filter query is not undefined

dlabrecq:
Fix isPlatformCost function, ensuring filter query is not undefined


dlabrecq (commented):
self

PR #2744, opened 2022-12-02
Jest warning

dlabrecq:
Split files for Jest, which will deoptimise styling if files exceed its max size. It's only a warning, but might help speed up the tests a bit.

Screen Shot 2022-12-01 at 7 00 00 PM


dlabrecq (commented):
self

PR #2756, opened 2022-12-06
Reset platform costs toggle

dlabrecq:
Reset platform costs toggle when group_by changes


dlabrecq (commented):
self

PR #2760, opened 2022-12-07
@redhat-cloud-services/frontend-components-utilities update

dlabrecq:
Update @redhat-cloud-services/frontend-components-utilities pkg


dlabrecq (commented):
self

PR #2763, opened 2022-12-07
Wrap "Cost distribution" wizard step with Unleash

dlabrecq:
The cost models wizard was updated to add a "Cost distribution" step via  https://issues.redhat.com/browse/COST-3145. However, the selections for "unallocated capacity" are commented out in the code.

This wraps the "Cost distribution" feature with this flag, so it can be available for testing.

https://issues.redhat.com/browse/COST-3319


dlabrecq (commented):
self

PR #2772, opened 2022-12-09
Export dialog warning

dlabrecq:
Radio buttons should use isChecked to avoid "uncontrolled" warning messages.

https://issues.redhat.com/browse/COST-3299


dlabrecq (commented):
self

PR #2773, opened 2022-12-09
Add cost_type for all AWS and "AWS on OCP" overview cards

dlabrecq:
Add cost_type for all AWS and "AWS on OCP" overview cards. This change adds the cost_type param for compute and storage APIs.

https://issues.redhat.com/browse/COST-3122


dlabrecq (commented):
self

PR #2776, opened 2022-12-12
cost model tables update

gitdallas:
[3230](https://issues.redhat.com/browse/COST-3230)

Migrate cost model and wizard pages to use the PatternFly composable table.


dlabrecq (commented):
Thanks for the PR. The cost model tables look ok, except the wizard's review step.

For the review step, I encountered an action button with no label for the price list table. I suspect the "remove" button shouldn't be available in the review step.

There is also some wrapping and truncation here. In the screenshot below; for example, the "Usage (core-hours) wraps onto the next line. And the "Description" and "Metric" table headers are truncated as "Des" and "Me".

I believe you can use modifier="nowrap" to solve the line wrapping. Not sure about the truncation. You might be able to use modifier="fitContent" to avoid truncating or a min-width?

wizard

dlabrecq:
If you're up for it, I found one more table in src/routes/costModels/costModel/sourceTable.tsx -- note that's not sourcesTable. If not, that's fine -- I can update that one later.

gitdallas:
> If you're up for it, I found one more table in src/routes/costModels/costModel/sourceTable.tsx -- note that's not sourcesTable. If not, that's fine -- I can update that one later.

that `<Table>` actually references `/costModel/table.tsx` which then uses SourcesTable

gitdallas:
@dlabrecq  - for the table modifications, after removing the actions column where necessary this is a comparison between the old:
image

and new:
image

the old tables wrap as well. something would need to happen if there is not enough space.. and the possible length of data can vary, especially with description.

dlabrecq (commented):
Looks like we need to update the priceListTable for the wizard as well.
I sorted the rate table by metric. When I attempted to delete the last row, I actually deleted the first row.

chrome-capture-2022-11-15

dlabrecq (approved):

PR #2777, opened 2022-12-12
Cost Model link in left menu doesn't work

dlabrecq:
Cost Model link in left menu doesn't work. Turns out that the path order in the router matters.


dlabrecq (commented):
self

PR #2781, opened 2022-12-13
Match new labels in details table

dlabrecq:
The APIs must also use `category=Platform` and `filter[category]=Platform`

https://issues.redhat.com/browse/COST-3335


dlabrecq (commented):
self

PR #2782, opened 2022-12-13
Filtering of cost models by source type doesn't work

dlabrecq:
Fixes a missing router property

https://issues.redhat.com/browse/COST-3340


dlabrecq (commented):
self

PR #2786, opened 2022-12-14
Use new project classifications

dlabrecq:
This updates the UI to use the new classifications provided by the cost APIs, instead of parsing project labels.
Also moved commonly used properties to `utils/props.ts`.

Summary of classifications:

- default for the projects that are kube- or openshift-
- category_<category_name> for the categories, so category_Platform for the Platform category
- unallocated for the Platform unallocated and Worker unallocated projects
- project for anything that does not fit any of the above criteria

https://issues.redhat.com/browse/COST-3347


dlabrecq (commented):
self

PR #2792, opened 2022-12-19
Add category filter to tags API

dlabrecq:
Adds category filter to tags API in OCP breakdown

https://issues.redhat.com/browse/COST-3346


dlabrecq (commented):
self

PR #2803, opened 2022-12-29
Fix cost models table test

dlabrecq:
The recent PatternFly table upgrade is generating the test error below. The containerized build is failing, so new PRs are blocked from merging.

Note that this is a temporary fix. Ideally, we would adjust the react-intl mock; however, the snapshots are changing during PR build tests.

Screen Shot 2022-12-29 at 11 57 42 AM


dlabrecq (commented):
self

PR #2813, opened 2023-01-07
Breakdown page doesn't show all "platform" clusters

dlabrecq:
This ensures the clusters card (shown in the breakdown page) shows all clusters for the "platform" feature.


dlabrecq (commented):
self

PR #2821, opened 2023-01-10
COST-3374: Update details breakdown query to use ID over label

chambridge:
* When RBAC is enabled the query fails as the user isn't given access permissions via user-friendly name
* It was decided to avoid doing a lookup that adds the alias or source name to forego and potential privilege escalation

Related to:
https://issues.redhat.com/browse/COST-3374


dlabrecq (approved):
Looks ok. 

I'll add an id prop (in another PR), so it's clear what we're using here.

PR #2849, opened 2023-01-19
Update page titles

dlabrecq:
Update page titles, ensuring all words start with uppercase letters

https://issues.redhat.com/browse/COST-3396


dlabrecq (commented):
self

PR #2865, opened 2023-01-26
OCI typeahead support

dlabrecq:
Enabled typeahead support for the OCI details page.

https://issues.redhat.com/browse/COST-3413


dlabrecq (commented):
self

PR #2889, opened 2023-01-31
"to skip this step" message

dlabrecq:
The wizard's "to skip this step" message is shown twice.

https://issues.redhat.com/browse/COST-3437


dlabrecq (commented):
self

PR #2901, opened 2023-02-07
Unleash for ROS

dlabrecq:
Want to add Unleash and routing for ROS.

- Create Unleash flag for stage and prod
- Add Unleash support to FeatureFlag component and Redux store
- Add routing paths for recommendation pages
- Add empty state message when user doesn't have access to pages

https://issues.redhat.com/browse/COST-3478


dlabrecq (commented):
self

PR #2902, opened 2023-02-07
Boiler plate for ROS APIs

dlabrecq:
Added support for ROS APIs. The recommendations API request will be based on OCP for now.

- Add support for API requests
- Add Redux support for API

https://issues.redhat.com/browse/COST-3481


dlabrecq (commented):
self

PR #2908, opened 2023-02-10
Boiler plate for ROS pages

dlabrecq:
Add boiler plate support for a recommendations page, based on OCP for now.
This will be available in the stage env only, under an Unleash flag.

https://issues.redhat.com/browse/COST-3477


dlabrecq (commented):
self

PR #2918, opened 2023-02-14
Missing tag link for OCI breakdown

dlabrecq:
The tag link is missing from the OCI breakdown page.

https://issues.redhat.com/browse/COST-3498


dlabrecq (commented):
self

PR #2919, opened 2023-02-14
Fetch user ID for Unleash

dlabrecq:
Ensure Unleash features are not invoked until we have the fetched the user ID from Insights.

https://issues.redhat.com/browse/COST-3499


dlabrecq (commented):
self

PR #2920, opened 2023-02-15
Insights Chrome session token

dlabrecq:
The Insights Chrome no longer appears to add the `cs_jwt` session token to local storage. As a result, the UI cannot properly clear local storage properties upon a new session. The local session is used to store user choices for currency, AWS cost type, and inactive sources alerts.

Instead of referencing the `cs_jwt` property in local storage, we can use `insights.chrome.auth.getToken()` instead.

https://issues.redhat.com/browse/COST-3500


dlabrecq (commented):
self

PR #2922, opened 2023-02-15
Use PatternFly's getCustomTheme function

dlabrecq:
Charts should use PatternFly's getCustomTheme function instead of mergeTheme. The mergeTheme function will be removed for the upcoming v5 breaking change release.

https://issues.redhat.com/browse/COST-3496


dlabrecq (commented):
self

PR #2935, opened 2023-02-21
Refactor to use Insights' shared router

dlabrecq:
Refactor to use Insights' shared router, formatting routes with a basename.

https://issues.redhat.com/browse/COST-3524


dlabrecq (commented):
self

PR #2938, opened 2023-02-22
OCP breakdown cluster name truncation

dlabrecq:
The OCP breakdown page shows a list of clusters, but some names are too long and must be truncated. It would be ideal if we could show a tooltip to display the full cluster name.

We also show a "more" link when the entire cluster string is greater than 50 characters. However, instead of 50 characters total, we can apply a limit to each cluster name. The cluster card is responsive, so cluster names will wrap onto the next line.

That said, max characters per cluster name should be 45 – assuming there is no whitespace. Otherwise, the cluster name may overlap the card.

Note: Max characters is 30 for this screenshot only
Screen Shot 2023-02-22 at 11 09 51 AM

https://issues.redhat.com/browse/COST-3533


dlabrecq (commented):
self

PR #2942, opened 2023-02-24
Disable table toolbar actions for no data

dlabrecq:
There are scenarios when table toolbar actions should be disabled. Including the bulk select, filter, include/exclude, export, manage columns, platform costs, and pagination.

1. While the page is loading, all toolbar actions are disabled
2. If the API returns no data, all toolbar actions are disabled
3. If there is no data due to filtering, only the filter actions remain enabled

https://issues.redhat.com/browse/COST-3537


dlabrecq (commented):
self

PR #2968, opened 2023-03-08
Incorrect alignment of Infra and Supp cost columns

dlabrecq:
The PatternFly composable table appears to have changed and broke our previous column alignments

https://issues.redhat.com/browse/COST-3568

Screenshot 2023-03-08 at 6 00 30 PM


dlabrecq (commented):
self

PR #2984, opened 2023-03-15
Rename ros as recommendations

dlabrecq:
- Rename ros as recommendations
- Fix recommendations header styles
- Refactor drawer components


dlabrecq (commented):
self

PR #2992, opened 2023-03-18
Fix CSS scope name

dlabrecq:
Fix CSS scope name to eliminate extra div tag / selector, which helps make ROS drawer sticky


dlabrecq (commented):
self

PR #3004, opened 2023-03-21
Convert to npm package manager

dlabrecq:
Convert from yarn to npm package manager in order to resolve peer dependency issues


dlabrecq (commented):
self

PR #3007, opened 2023-03-21
Use PatternFly sticky table props

dlabrecq:
Refactor explorer table to use PatternFly sticky column props

https://issues.redhat.com/browse/COST-3633

Screenshot 2023-03-21 at 7 51 30 PM


dlabrecq (commented):
self

PR #3018, opened 2023-03-24
Update node

dlabrecq:
Update node to the latest LTS release and `jammy` dist (Ubuntu 22.04)

Note: The Node.js 18 is compatible only with Linux distributions based on glibc 2.28 or later.
In order to use NodeJS 18, install it on a server with supported OS, e.g. Debian 10, RHEL 8, Ubuntu 20.04.

See https://support.plesk.com/hc/en-us/articles/5517433733394--node-or-npm-CLI-commands-on-a-Plesk-server-fail-version-GLIBC-2-28-not-found-required-by-opt-plesk-node-18-bin-node-


dlabrecq (commented):
self

PR #3021, opened 2023-03-27
Remove trailing comma from filter field input

dlabrecq:
Removed trailing comma and ampersand from filter field input

https://issues.redhat.com/browse/COST-3641


dlabrecq (commented):
@dlabrecq

PR #3022, opened 2023-03-27
Revert fix for COST-1131

dlabrecq:
Revert fix for COST-1131 -- we shouldn't omit filters


dlabrecq (commented):
self

PR #3034, opened 2023-03-30
Omit filters associated with the current group_by

dlabrecq:
Omit filters associated with the current group_by


dlabrecq (commented):
self

PR #3039, opened 2023-04-03
Ensure correct export date range

dlabrecq:
Ensure correct date range is used for previous month export

https://issues.redhat.com/browse/COST-3659


dlabrecq (commented):
self

PR #3041, opened 2023-04-03
Remove extra auth interceptor call.

chambridge:
* remove extra auth interceptor call adding additional Bearer token.


dlabrecq (approved):
I removed the unused import, which was causing the lint test to fail.

FYI, when Boaz introduced `insightsAuthInterceptor` via https://github.com/project-koku/koku-ui/pull/1250, I believe the intention was so we could skip testing for `window.insights`.

I'll add that back via a different PR.

PR #3042, opened 2023-04-03
Ensure chrome user exists before making API requests

dlabrecq:
When Boaz introduced `insightsAuthInterceptor` via https://github.com/project-koku/koku-ui/pull/1250, the intention was to skip testing for window.insights.

Now that `insightsAuthInterceptor` has been removed via https://github.com/project-koku/koku-ui/pull/3041, I've restored those tests in this PR. This will ensure the insights `user` exists before making requests.

Closes https://issues.redhat.com/browse/COST-3671


dlabrecq (commented):
self

PR #3045, opened 2023-04-04
Revert previous API update

dlabrecq:
Reverting previous API update to call insights.chrome.auth.getUser. No longer seems necessary to obtain the user prior to making API requests.


dlabrecq (commented):
self

PR #3046, opened 2023-04-04
Ensure platform project retains filters

dlabrecq:
There are multiple API requests in the breakdown page where we need to ensure filters are not cleared for the OCP platform project.


dlabrecq (commented):
self

PR #3049, opened 2023-04-06
ROS filtering and pagination

dlabrecq:
Added ROS filtering and pagination to optimizations page

- Updated IT optimization icon
- Wired filtering and pagination to ROS APIs
- Create a separate optimizations component for the breakdown page


dlabrecq (commented):
self

PR #3051, opened 2023-04-06
PatternFly progress bar fix

dlabrecq:
PatternFly progress bar cannot handle formatted locale values.

https://issues.redhat.com/browse/COST-3673


dlabrecq (commented):
self

PR #3053, opened 2023-04-06
Update page title component to use all page titles

dlabrecq:
Currently, only the overview and cost explorer have unique page titles

- The overview page title reads 'Overview - Cost Management | OpenShift'
- The explorer is 'Cost Explorer - Cost Management | OpenShift'
- All other pages are 'Cost Management | OpenShift'

I confirmed with Natalie that all pages should have unique titles


dlabrecq (commented):
self

PR #3058, opened 2023-04-08
Added empty state for breakdown tabs

dlabrecq:
When navigating from the Optimizations drawer to the breakdown page, the project may not exist for Cost Management APIs. In this case, the projects cost is zero and the "Cost overview" and "Historical data" tabs don't have data to show.

I used the empty state we show in the OpenShift details page when there is no data. I realize this is fake data, but feel we should have something in place. However, it is possible that the Cost and ROS APIs are not in sync when data is updated, sources are added / removed, etc.


dlabrecq (commented):
self

PR #3060, opened 2023-04-10
AWS org unit filter workaround

dlabrecq:
Added a workaround for AWS org unit filters

https://issues.redhat.com/browse/COST-3680


dlabrecq (commented):
self

PR #3063, opened 2023-04-10
Ensure currency is not stored as undefined

dlabrecq:
If account-settings returns `currency=undefined`, ensure that is not stored in local storage.

https://issues.redhat.com/browse/COST-3683


dlabrecq (commented):
self

PR #3065, opened 2023-04-11
Cost distribution

dlabrecq:
Add "Cost distribution" to cost models cards

https://issues.redhat.com/browse/COST-3249


dlabrecq (commented):
self

PR #3071, opened 2023-04-11
Adjusted query params for for "All projects" modal

dlabrecq:
Adjusted query params for for "All projects" modal

https://issues.redhat.com/browse/COST-3684


dlabrecq (commented):
self

PR #3074, opened 2023-04-11
Remove unused i18n messages

dlabrecq:
There are more than a few unused i18n messages, which could be removed. Wrote a script to check what keys are in use in the source code.

https://issues.redhat.com/browse/COST-3701


dlabrecq (commented):
self

PR #3077, opened 2023-04-12
Added function to check for empty ROS objects

dlabrecq:
Added function to help determine if config and variation are empty objects


dlabrecq (commented):
self

PR #3079, opened 2023-04-12
Rename workload in filter dropdown

dlabrecq:
Rename workload as workload name in filter dropdown


dlabrecq (commented):
self

PR #3083, opened 2023-04-14
Added cost distributed chart

dlabrecq:
- Added cost distribution chart to breakdown page
- Rename workers as worker
- Clean up local storage


dlabrecq (commented):
self

PR #3096, opened 2023-04-18
Check fetch status before dispatch

dlabrecq:
Check fetch status before dispatch

https://issues.redhat.com/browse/COST-3729


dlabrecq (commented):
self

PR #3104, opened 2023-04-18
Add distributed cost as an order by option

dlabrecq:
Add distributed cost as an order by option

https://issues.redhat.com/browse/COST-3722


dlabrecq (commented):
self

PR #3105, opened 2023-04-19
Add dropdown filter for workload types

dlabrecq:
Want to create a dropdown filter to select pre-canned workload types.

https://issues.redhat.com/browse/COST-3734

example


dlabrecq (commented):
self

PR #3112, opened 2023-04-20
Cost distribution

dlabrecq:
Add default costDistribution for breakdown header

https://issues.redhat.com/browse/COST-3740


dlabrecq (commented):
self

PR #3126, opened 2023-04-25
Reset delta when group by is changed

dlabrecq:
Reset delta when group by is changed. Also ensure we don't set order_by and/or overhead costs unless the group_by is 'project'


dlabrecq (commented):
self

PR #3128, opened 2023-04-26
OCP details cluster costs

dlabrecq:
- Fix report item value types for cost distribution
- Use ref to define table's stickyLeftOffset

https://issues.redhat.com/browse/COST-3756


dlabrecq (commented):
self

PR #3132, opened 2023-04-27
Remove unused app navigation

dlabrecq:
Remove unused app navigation from app.tsx. This will match Consoledot's starter app.


dlabrecq (commented):
self

PR #3134, opened 2023-04-27
Rename global store variable

dlabrecq:
Rename global store variable to help ensure it's not overridden by HCS

https://issues.redhat.com/browse/RHCLOUD-25573


dlabrecq (commented):
self

PR #3143, opened 2023-05-02
Initial settings page

dlabrecq:
- Add Unleash flags for settings page
- Add route for settings page
- Add initial settings page header and tabs

https://issues.redhat.com/browse/COST-3307


dlabrecq (commented):
self

PR #3147, opened 2023-05-03
Cost models tab

dlabrecq:
Added cost model features to new settings page

- Preserve history by moving cost models to settings directory
- Added cost models to deprecated directory
- Added cost models tab content
- Converted tables to be compact
- Updated details, cost explorer, and optimizations to use compact tables
- Fixed sticky column width
- Fixed cost distribution for cost explorer


dlabrecq (commented):
se;f

PR #3173, opened 2023-05-10
[COST-3554] Use new field for amortized cost

samdoran:
The value in `savingsplan_effective_cost` does not accurately reflect the amortized cost as reported by AWS. There is a new field that calculates the correct amortized cost the same way that AWS does. Use this new field when amortized cost is selected.

### DIFF
--- a/src/routes/components/costType/costType.tsx
+++ b/src/routes/components/costType/costType.tsx
@@ -43,7 +43,7 @@ type CostTypeProps = CostTypeOwnProps & CostTypeDispatchProps & CostTypeStatePro
 
 // eslint-disable-next-line no-shadow
 export const enum CostTypes {
-  amortized = 'savingsplan_effective_cost',
+  amortized = 'calculated_amortized_cost',
   blended = 'blended_cost',
   unblended = 'unblended_cost',
 }


samdoran:
Turns out there was a minor UI change needed after all. Sorry for any confusion. I don't think this should be a huge change (famous last words).

dlabrecq (changes requested):
Is the API change available in stage?

Multiple API requests error with "400 Bad Request".

{"cost_type":["\"calculated_amortized_cost\" is not a valid choice."]}

**API request example**
`api/cost-management/v1/reports/aws/costs/?cost_type=calculated_amortized_cost&currency=USD&filter[resolution]=daily&filter[time_scope_units]=month&filter[time_scope_value]=-2`

**Browser inspector**
Screenshot 2023-05-10 at 10 33 26 PM

dlabrecq:
@samdoran Why does the API support both `calculated_amortized_cost` and `savingsplan_effective_cost`?

Why not just fix the existing `savingsplan_effective_cost` param if that's not accurate?

samdoran:
API changes haven't been merged yet. https://github.com/project-koku/koku/pull/4365

I'll switch this to a draft until after that is merged.

samdoran:
> Why does the API support both calculated_amortized_cost and savingsplan_effective_cost?

That's for backwards compatibility for anyone using the API directly and requesting `savingsplan_effective_cost`.

> Why not just fix the existing savingsplan_effective_cost param if that's not accurate?

That value comes directly from AWS. They don't provide a value in the cost report data for amortized cost that matches how they calculate it, so we have to calculate it ourselves based on existing data and store that. We could overwrite `savingsplan_effective_cost` but that may confuse someone who notices the value from the cost report is different than what our app reports.

samdoran:
Turns out I was wrong about where this setting is used: this doesn't change the value stored in settings but rather the value used in the Show cost as drop down menu.

https://github.com/project-koku/koku/pull/4365 is deployed to stage. We can merge this in order to test it in stage.

This needs to be merged to prod _after_ https://github.com/project-koku/koku/pull/4365 is deployed to prod.

dlabrecq (approved):
Tested UI locally and LGTM

samdoran:
Thank you!

PR #3574, opened 2023-12-12
Update Redux to use latest packages

dlabrecq:
https://issues.redhat.com/browse/COST-4511


dlabrecq:
@dependabot ignore this major version

PR #3772, opened 2024-04-09
CI - Reenable coverage and update workflow

samdoran:
- Rename the pull_request workflow to CI
- Run jobs on pushes to important branches not just on PRs
- Add codecov action
- Update test command to use jest flags
- Improve Markdown formatting
- Update missing links
- Change CI status badge to GitHub actions
- Add coverage badge
- Update docker compose commands

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,16 +27,16 @@ jobs:
         node-version: [18]
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Setup Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
 
       - name: Cache node modules
         id: cache-npm
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         env:
           cache-name: cache-node-modules
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,12 +19,48 @@ env:
   BRANCH: ${{ github.base_ref }}
 
 jobs:
-  build:
-    name: koku-ui build
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
+
+      - name: Cache node modules
+        id: cache-npm
+        uses: actions/cache@v4
+        env:
+          cache-name: cache-node-modules
+        with:
+          # npm cache files are stored in `~/.npm` on Linux/macOS
+          path: ~/.npm
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Lint
+        run: npm run lint
+
+  units:
+    name: Units - ${{ matrix.node-version }}
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        node-version: [18]
+        node-version:
+          - '18'
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -28,9 +29,9 @@ sudo bash scripts/patch-etc-hosts.sh
 
 ## Getting Started
 1. Install requirements listed above.
-2. Setup /etc/hosts entries listed above. 
-3. Clone the repository, and open a terminal in the base of this project.
-4. Run the command `npm install` to install all the dependencies.
+1. Setup `/etc/hosts` entries listed above.
+1. Clone the repository, and open a terminal in the base of this project.
+1. Run the command `npm install` to install all the dependencies.
```
dlabrecq, line 34: Please fix the numbering

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
```
dlabrecq, line 33: Why do we need separate jobs for lint and units? It runs thru the many of the same steps twice

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
+name: CI
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - prod-*
+      - stable-*
```
dlabrecq, line 8: Do we need this to run on push?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
...
+        run: npm run lint
+
+  units:
+    name: Units - ${{ matrix.node-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version:
+          - '18'
```
dlabrecq, line 62: Don't want to hard code this version, see ${{ matrix.node-version }}

dlabrecq (changes requested):
We're still relying on pull_request.xml to build PRs because the pipeline fails so often.
I was just expecting to re-add the "Upload coverage report" step at the bottom of that file.

README.md
samdoran: This is Markdown syntax for on ordered list. The correct numbers are used when the page is rendered. Makes reordering steps much less of a hassle.

.github/workflows/ci.yml
samdoran: Separate jobs lets them run concurrently, reducing overall runtime. There are some duplicate steps, but that's because the setup needs to happen on each runner.

.github/workflows/ci.yml
samdoran: Right now we don't have any CI runs on the code _after_ it is merged into `main` or other branches. This would let us know the status of the code in those branches not just the PR branches.

.github/workflows/ci.yml
samdoran: This is the definition of `matrix.node-version`. It looks funny because it is reference in thae `name` field, but that's how `matrix` works.

.github/workflows/ci.yml
samdoran: The other reason for this is integrity of the CI status badge. With a test run against the stable branches, the CI status badge would only indicate that the last PR run was successful, not that the code in the stables branches was tested successfully.

samdoran:
> We're still relying on pull_request.xml to build PRs because the pipeline fails so often.

Do you mean the "ci.ext.devshift.net PR build" check fails often?

I renamed `pull_request.yml` to `ci.yml` . It's doing the same checks, just optimized a bit and running against stable branches as well.

The file can be named whatever we want.

.github/workflows/ci.yml
samdoran: We could avoid the setup steps by using our own container image for CI runs. That's a much more reliable (and faster) way to run tests, but requires maintaining the CI image(s).

README.md
samdoran: I'm happy to go back to using hard coded numbers, though.

samdoran:
Speaking of ci.ext.devshift.net, that check is currently failing due to a bug (https://github.com/RedHatInsights/insights-frontend-builder-common/pull/189). I'll have to change the branch name for this PR to work around it, which means this will get closed.

I can open one PR per commit in this branch so it's easier to review.

PR #3794, opened 2024-04-11
Consolidate CI workflows

samdoran:
Use one workflow with multilpe jobs that run concurrently.
Since we only ever need to test with one version of Node.js, remove matrix strategy.


dlabrecq (approved):
LGTM

PR #5421, opened 2026-07-12
COST-7658: Enable on-prem Optimizations box-plot Unleash flag by default

jkilzi:
## Summary
- On-prem has no Unleash proxy, so `cost-management.koku-ui-ros.box-plot` stayed off and Optimizations utilization charts never rendered.
- Introduce `DEFAULT_ONPREM_UNLEASH_FLAGS` / `resolveOnpremUnleashFlags()` and bake that default through on-prem webpack `DefinePlugin` + the Unleash stub (minimal scope: box-plot only).
- Explicit `ONPREM_UNLEASH_FLAGS` still overrides the default list.

## Test plan
- [x] Unit: `npm test -w @koku-ui/onprem-cloud-deps`
- [x] Seed ROS data (`setup-test-data.sh --scenario ros --days 15 --no-cleanup`)
- [x] `npm run start:onprem:auth` → Optimizations list → breakdown shows **CPU utilization** and **Memory utilization** boxplots with recommended request/limit lines
- [x] Follow-up (after Dan ACK): expand defaults to other always-on SaaS Unleash flags (exclude `namespace` until prod; exclude account-gated `aws-ec2-instances`)

Resolves COST-7658 (on-prem). SaaS already works with the live Unleash flag.

Made with [Cursor](https://cursor.com)


libs/onprem-cloud-deps/src/unleash/proxy-client-react.ts
```diff
@@ -1,10 +1,12 @@
...
 
+import { resolveOnpremUnleashFlags } from './defaultFlags';
+
 const readUnleashFlagsEnv = (): string => {
   if (typeof process === 'undefined') {
-    return '';
+    return resolveOnpremUnleashFlags(undefined);
   }
-  return process.env.ONPREM_UNLEASH_FLAGS ?? '';
+  return resolveOnpremUnleashFlags(process.env.ONPREM_UNLEASH_FLAGS);
```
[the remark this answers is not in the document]
y0n1: Fixed: switched to try/catch so DefinePlugin-injected `process.env.ONPREM_UNLEASH_FLAGS` literals are evaluated in the browser.

libs/onprem-cloud-deps/package.json
```diff
@@ -28,7 +28,7 @@
     "lint": "npm-run-all lint:*",
     "lint:ts": "eslint src",
     "lint:ts:fix": "eslint src --fix",
-    "test": "",
+    "test": "node --experimental-strip-types --test 'test/**/*.test.ts'",
```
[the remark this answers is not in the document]
y0n1: Fixed: monorepo already requires Node `>=22.20.0` (same as `@koku-ui/token-refresher` tests). Updated `apps/koku-ui-onprem/README.md` and added matching `engines` on `@koku-ui/onprem-cloud-deps`.

libs/onprem-cloud-deps/src/unleash/defaultFlags.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * On-prem builds have no live Unleash proxy. Flags listed here are enabled by
+ * default when `ONPREM_UNLEASH_FLAGS` is unset/empty (webpack DefinePlugin + stub).
+ *
+ * Expand carefully after confirming each flag is always-on in SaaS prod.
+ * See COST-7658.
+ */
+export const BOX_PLOT_FLAG = 'cost-management.koku-ui-ros.box-plot';
```
dlabrecq: I don't believe should redefine Unleash flags here. When we remove the flag, we'll have to remember this also appears in the onprem-cloud-deps package.

Can we just add the flag to webpack-onprem.config file in koku-ui-ros? That way, it's just the ROS app we have to track

apps/koku-ui-hccm/webpack-onprem.config.ts
```diff
@@ -7,6 +7,8 @@ import TerserJSPlugin from 'terser-webpack-plugin';
 import type { Configuration } from 'webpack';
 import { DefinePlugin } from 'webpack';
 
+import { resolveOnpremUnleashFlags } from '../../libs/onprem-cloud-deps/src/unleash/defaultFlags';
```
dlabrecq: This will import the ROS box plot flag, which is not used in koku-ui-hccm.

apps/koku-ui-ros/webpack-onprem.config.ts
```diff
@@ -7,6 +7,8 @@ import TerserJSPlugin from 'terser-webpack-plugin';
 import type { Configuration } from 'webpack';
 import { DefinePlugin } from 'webpack';
 
+import { resolveOnpremUnleashFlags } from '../../libs/onprem-cloud-deps/src/unleash/defaultFlags';
```
dlabrecq: This will import any Unleash flags for the koku-ui-hccm app, which would be unused here.

apps/koku-ui-onprem/webpack.config.ts
```diff
@@ -11,6 +11,8 @@ import type { Configuration } from 'webpack';
 import { container, DefinePlugin } from 'webpack';
 import type { Configuration as WebpackDevServerConfiguration } from 'webpack-dev-server';
 
+import { resolveOnpremUnleashFlags } from '../../libs/onprem-cloud-deps/src/unleash/defaultFlags';
```
dlabrecq: It's not clear why we would add Unleash flags for koku-ui-hccm and koku-ui-ros here? I believe we want to keep these flags separate

dlabrecq (changes requested):

libs/onprem-cloud-deps/src/unleash/defaultFlags.ts
jkilzi: Good call — dropped the shared defaults from `onprem-cloud-deps`. `cost-management.koku-ui-ros.box-plot` is now defaulted only in `apps/koku-ui-ros/webpack-onprem.config.ts`, so flag removal stays scoped to the ROS app.

apps/koku-ui-hccm/webpack-onprem.config.ts
jkilzi: Agreed — HCCM shouldn’t get the ROS box-plot default. Reverted this to the previous `ONPREM_UNLEASH_FLAGS` passthrough.

apps/koku-ui-ros/webpack-onprem.config.ts
jkilzi: Agreed on keeping app flags separate. Removed the shared helper and inlined the ROS-only default here (no HCCM flags).

apps/koku-ui-onprem/webpack.config.ts
jkilzi: Agreed — the shell shouldn’t bake HCCM/ROS Unleash defaults. Reverted; ROS remotes get their own default from `webpack-onprem.config.ts`.

dlabrecq (approved):

PR #5428, opened 2026-07-13
Update webpack-dev-server dependency

dlabrecq:
`webpack-dev-server` 6 upgrades `http-proxy-middleware` to v4, which drops top-level event handlers like `onProxyReq`. Events now go under an on object, so `devServerProxy.ts` must be updated.

### DIFF
--- a/apps/koku-ui-onprem/package.json
+++ b/apps/koku-ui-onprem/package.json
@@ -58,7 +58,6 @@
     "style-loader": "^4.0.0",
     "ts-loader": "^9.6.2",
     "ts-node": "^10.9.2",
-    "typescript": "^5.9.3",
-    "webpack-dev-server": "^5.2.5"
+    "typescript": "^5.9.3"
   }
 }
--- a/apps/koku-ui-onprem/webpack.config.ts
+++ b/apps/koku-ui-onprem/webpack.config.ts
@@ -146,18 +146,22 @@ const config: Configuration & {
     },
     setupMiddlewares,
     proxy: [
-      {
-        context: ['/api/cost-management/v1'],
-        target: process.env.API_PROXY_URL,
-        changeOrigin: true,
-        secure: false,
-        pathRewrite: { '^/api/cost-management/v1': '' },
-        // Pass both HTTP and HTTPS traffic through the PAC agent
-        ...(pacAgent && { agent: pacAgent }),
-        // In oauth2-proxy mode proxyHeaders is undefined — the Authorization
-        // header injected by oauth2-proxy is forwarded as-is to the gateway.
-        ...(proxyHeaders && { headers: proxyHeaders }),
-      },
+      ...(process.env.API_PROXY_URL
+        ? [
+            {
+              context: ['/api/cost-management/v1'],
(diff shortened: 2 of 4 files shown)


jkilzi (changes requested):
Hey Dan — I took a closer look at this one (and also how it interacts with the planned COST-7903 TokenRefresher removal).

**Requesting changes** on the `webpack-dev-server` bump as written.

The `onProxyReq` → `on.proxyReq` migration matches WDS 6 / http-proxy-middleware v4, but in this monorepo `webpack serve` for `koku-ui-onprem` does **not** load the nested `webpack-dev-server@6` install.

What actually happens after this PR:
- FEC (`frontend-components-config`) pins `webpack-dev-server@5.2.3` → that copy is hoisted to the **repo root**
- `@koku-ui/koku-ui-onprem` / `@koku-ui/token-refresher` get a **nested** `6.0.0`
- `start` uses `../../node_modules/.bin/webpack`, so `@webpack-cli/serve` does `require("webpack-dev-server")` from the **root** package → **5.2.3** (hpm v2), never the nested 6

With hpm v2, handlers are read from top-level `onProxyReq`, not `on.proxyReq`. Feeding it the new shape returns no `proxyReq` handler, so the Keycloak `createDevServerProxy` Authorization injection silently never fires on the default local Keycloak path (`start:onprem:dev`).

CI stays green because this is a `webpack serve`-only gap.

**Fix direction:** same pattern as #5429 for webpack — hoist / override a single root `webpack-dev-server@6` (don’t rely on a nested workspace install), or otherwise make webpack-cli resolve 6. README still mentions `onProxyReq` too.

---

**Separately — COST-7903 timing**

We’re planning to remove `libs/token-refresher` and `start:onprem:dev` (COST-7903). I simulated merging this PR onto that removal:

- modify/delete conflicts on `libs/token-refresher/package.json` and `devServerProxy.ts`
- the only consumer of the v6-only `on.proxyReq` API is exactly what COST-7903 deletes

So if COST-7903 lands first, most of this PR becomes **superseded** (the broken path is gone), and the remaining `koku-ui-onprem` `webpack-dev-server: ^6` bump is still ineffective at runtime unless paired with a root hoist/override.

Happy to sync on order: either (a) fix the root hoist here if we still need TokenRefresher for a bit, or (b) close/supersede this once COST-7903 lands and revisit a WDS 6 bump only if we still want it for other reasons.

dlabrecq:
I added an override similar to webpack in the root package. And fixed cypress tests by only adding proxy entries when API_PROXY_URL is set

jkilzi (approved):
Verified in a worktree — `webpack-dev-server` now resolves to the root `6.0.0` install via `@webpack-cli/serve`'s resolution (confirmed with `require.resolve`), and `webpack serve` boots cleanly with it. The Authorization-header proxy injection also no longer depends on the `onProxyReq`/`on.proxyReq` event shape at all (it's set statically via the proxy `headers` option), so the hpm v2→v4 concern doesn't apply here anymore. CI is green across the board. LGTM.

jkilzi:
Let's merge this and sync stage-onprem with main.

PR #5429, opened 2026-07-13
Update webpack dependency

dlabrecq:
The koku-ui-onprem build fails when updating the webpack dependency. This is due to duplicate webpack installations -- FEC uses an older webpack dependency.

"TypeScript treats these as different, incompatible types, even though both are webpack 5.x.

In an npm workspaces monorepo, webpack should be deduplicated to a single version at the root, not pinned in one app only.

When webpack is pinned to 5.108.4 in apps/koku-ui-onprem/package.json, npm installs it locally under that workspace. Shared plugins stay hoisted at the repo root against the older webpack." -- Cursor

### DIFF
--- a/apps/koku-ui-onprem/package.json
+++ b/apps/koku-ui-onprem/package.json
@@ -60,7 +60,6 @@
     "ts-loader": "^9.6.2",
     "ts-node": "^10.9.2",
     "typescript": "^5.9.3",
-    "webpack": "5.105.0",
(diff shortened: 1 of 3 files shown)


jkilzi (approved):
Verified locally: webpack is correctly hoisted to 5.108.4 via root + overrides (single deduped install), and `koku-ui-onprem` / `koku-ui-hccm` / `koku-ui-ros` production builds succeed.

LGTM.

PR #5437, opened 2026-07-16
Refactor Unleash flags for webpack-onprem.config.ts

dlabrecq:
Refactor Unleash flags for use with webpack-onprem.config.ts


jkilzi (approved):
LGTM — clean split of Unleash flag names into `FeatureToggleType` so webpack-onprem can share them, sensible on-prem defaults (including efficiency), and solid unit coverage. Gemini’s earlier duplicate-export note looks addressed by the rename/split.

This should unblock the Optimizations `h1` path on on-prem by enabling the efficiency toggle at build time.

PR #5495, opened 2026-07-28
fix(COST-7661): retarget on-prem Integrations links to Settings

jkilzi:
## Summary

- On-prem source links from OpenShift Integrations status (and related call sites) pointed at SaaS `/settings/integrations/detail/<id>`, which is outside the Cost Management mount and rendered a **blank page** ([COST-7661](https://issues.redhat.com/browse/COST-7661)).
- Retarget those links to Settings (`/openshift/cost-management/settings`) so the page renders. SaaS / Platform Integrations URLs are unchanged.
- **Out of scope:** path-style integration **detail** deep-link (c.rh.c `/settings/integrations/detail/<id>`). Deferred to [COST-7441](https://issues.redhat.com/browse/COST-7441) when Sources is federated as an MFE.

## Test plan

- [x] OpenShift details → Integrations status → source link **href** is `/openshift/cost-management/settings` (not `/settings/integrations/detail/…`)
- [x] Click → Settings loads with content (not blank)
- [x] Optional: Integrations tab lists sources; detail for the clicked source is **not** required here (COST-7441)
- [ ] SaaS: links still target Platform Integrations detail
- [x] Unit tests: `sourceLink`

## Notes for reviewers

Dropped the earlier `?source=` / Settings URL mutation and tabs refactor from review feedback. Verified on-prem with `quay.io/jkilzi/koku-ui-onprem:cost-7661-rc3`.


apps/koku-ui-hccm/src/routes/settings/settingsSourceSearch.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * On-prem Settings href for Sources list (`?source`) or detail (`?source=<uuid>`).
+ */
+export function buildOnPremSettingsSourceHref(settingsPath: string, uuid?: string | null): string {
+  if (uuid) {
+    return `${settingsPath}?${SETTINGS_SOURCE_PARAM}=${encodeURIComponent(uuid)}`;
+  }
+  return `${settingsPath}?${SETTINGS_SOURCE_PARAM}`;
+}
```
dlabrecq: This file would be better located with the SourceLink component. For example, routes/details/components/providerStatus/utils or in the SourceLink component itself.

apps/koku-ui-hccm/src/routes/settings/settings.tsx
```diff
@@ -111,66 +132,58 @@ const Settings: React.FC<SettingsProps> = () => {
...
-      ...(isSettingsSourcesTabEnabled
-        ? [
-            {
-              contentRef: React.createRef(),
-              tab: SettingsTab.sources,
-            },
-          ]
-        : []),
-    ];
-    return availableTabs;
```
dlabrecq, line 171: Why is the tabs functionality being rewritten? Please drop the tabs/ref refactor and `findTabIndexById` unless there’s a concrete bug they fix.

apps/koku-ui-hccm/src/routes/settings/settings.tsx
```diff
@@ -276,14 +289,23 @@ const Settings: React.FC<SettingsProps> = () => {
...
+    if (leavingSources && searchParams.has(SETTINGS_SOURCE_PARAM)) {
+      setSearchParams(
+        prev => {
+          const next = new URLSearchParams(prev);
+          next.delete(SETTINGS_SOURCE_PARAM);
+          return next;
+        },
+        { replace: true }
+      );
+    }
```
dlabrecq: This code is run every time a tab is clicked regardless if it's SaaS or onprem. I don't believe we should have to alter the settings page URL here.

That said, Stefan designed the price list feature so that when users drill down to a specific price list, it has its own page. The cost model breakdown behaves the same way. I believe the integrations breakdown should mimic these two settings features. This would eliminate the need to alter the settings URL and the breadcrumb appearance (for the breakdown) would be consistent with our other settings pages.

dlabrecq (changes requested):

jkilzi:
Thanks for the review — agreed on the direction after looking at c.rh.c Sources (Red Hat tab → `/settings/integrations/detail/<id>`).

**Focus of this PR (COST-7661):** fix the **blank page** only. On-prem links were pointing at the SaaS Integrations detail path, which is outside the Cost Management mount. I now retarget those links to Settings (`/openshift/cost-management/settings`) so the page actually renders. No Settings URL mutation, no `?source=` deep-link, and the earlier tabs/ref refactor is dropped.

**Deferred to COST-7441:** path-style integration **detail** drill-down (same model as price list / cost models, and as SaaS Sources today). That’s the right place to preserve Integrations breakdown routing once Sources is federated as an MFE.

Happy to adjust further if you want anything else called out in the PR description.

dlabrecq (approved):

PR #5557, opened 2026-08-09
Fix build-container step

dlabrecq:
Hermeto only downloads packages that have a resolved URL; without that it treats them as local file: deps and skips them, so npm ci fails offline. Added resolved URLs for packages Hermeto was treating as local file.

Note: if package-lock file is regenerated, those resolved backfills can be lost, and the hermetic build can fail again depending on how you regenerate.

To regenerate:
rm package-lock.json && npm install with current npm 11 (omit-lockfile-registry-resolved=false)

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+# Hermeto/Konflux hermetic builds need resolved URLs in package-lock.json.
+omit-lockfile-registry-resolved=false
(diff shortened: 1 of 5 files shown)


dlabrecq:
Closing in favor of https://github.com/project-koku/koku-ui/pull/5558

PR #5597, opened 2026-08-18
fix: use single onprem-candidate branch for koku-ui-onprem builds

ydayagi:
Replace the two-branch pattern (stage-onprem + prod-onprem) with a single onprem-candidate branch. Prod releases use the same snapshot via manual trigger in Konflux UI.

## Summary by Sourcery

Use the single `onprem-candidate` branch as the source for Koku UI on-premise builds and validation.

Enhancements:
- Consolidate on-premise build and pull-request validation workflows around the single `onprem-candidate` branch.

CI:
- Update Cypress and Tekton automation to target `onprem-candidate` instead of separate stage and production branches.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -4,8 +4,7 @@ on:
   pull_request:
     branches:
       - main
-      - stage-onprem
-      - prod-onprem
+      - onprem-candidate
 
 env:
   NODEJS_VERSION: '22'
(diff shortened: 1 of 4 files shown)


dlabrecq (changes requested):
There is another reference to stage-onprem in `.github/workflows/koku-ui-onprem-push-stage.yml`. Can you please update that as well and rename as `koku-ui-onprem-push.yml`

There's also a reference in `.github/workflows/cypress.yml`

dlabrecq (approved):

PR #5717, opened 2026-09-18
COST-8227: Use /currency/ for page currency dropdowns

jvcsizilio:
## Summary
- Point `SettingsType.currency` at `currency/` so Overview / Details / Explorer dropdowns work for view-only users (no settings RBAC).
- Keep `settings/currency/` for the Exchange rates settings page via `SettingsType.currencySettings`.
- Display tab still uses `account-settings/currency/` (unchanged).

## Test plan
- [ ] View-only user: Overview Currency dropdown works; Network shows `GET .../currency/` → 200
- [ ] Settings admin: Exchange rates still loads `settings/currency/`
- [ ] Display tab still uses account-settings for default currency

https://redhat.atlassian.net/browse/COST-8227


dlabrecq (approved):
Using the /currency API is fine, but need to make a tweak. While using /settings/currency, I added `filter%5Benabled%5D=true` to show just the enabled currencies. So, need to remove that.

There's also some getCurrencyOptions logic that may need to be refactored.

I'll follow up with another PR

